PR #9, opened 2016-12-20

ofear9:


07_cmd_lab/cmdlab.txt
```diff
@@ -1,13 +1,44 @@
...
+  
+  
+
+2. View the file /etc/passwd on your machine with the command "cat /etc/passwd". Then read about the file in "man 5 passwd". 
+What information is held in the file? What are the fields in each line?
+
+contains one line for each user account, with seven fields delimited by colons (“:”). These
+ +       fields are:
+ +       ·   login name
+ +       ·   optional encrypted password
```
vitalipe, line 24: just as a side note: modern linux/unix systems storing password hashed in ```/etc/shadow```.. 

so passwords here are ignored (it's for backward compatibility)

07_cmd_lab/cmdlab.txt
```diff
@@ -1,13 +1,44 @@
...
+ +       ·   user home directory
+ +       ·   optional user command interpreter
+
+
+
+
+3. In Unix, the /etc folder holds configuration files. 
+Display all files in "/etc" folder ordered by file size (Hint: search in "man ls")
+
+ls -sort -S /etc
```
vitalipe, line 37: ha! :smiley:  there is a subtle "bug" here (and I'm kinda glad you made it!) here let me attempt to explain:

what happens here is this:

``ls -s -o -r -r -S``

why? because there is a nice feature with command flags, it's possible to combine them, for example:

```ls -lh``` is the same as ```ls -l -h```

world flags are prefixed with "--", e,g: ```ls --all``` 

it's kinda awesome that ls has all  flags in the string "sort" :smile:

vitalipe:
@ofear9 - I'm closing this , well done :)

PR #18, opened 2016-12-24

dovron:


12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,44 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
   5. All files with extension ".c" that are in a folder named "src"
   Answer:
+	
+	ls ./src/*.c
 
   6. All files with extension ".c" under any folder
   Answer:
 
+    find . -name '*.c'
+	ls -R | grep '.*[.]c'
```
vitalipe, line 35: wow nice :) really liked your solution! here are a few notes:

this command ``` ls -R | grep '.*[.]c ```  
will also match the file ```fake.c.other```
to fix this add a ```$``` to the end: ``` ls -R | 'grep '.*[.]c$'``` 
(this is a regex thing...)

also there is a cool and almost  hidden feature of bash (4 and up) called "globstar" that allows you to create recursive globs, like this:
``` ls **/*.c # will match any .c file under current dir```

just remember that you need to enable it first: ``` shopt -s globstar ```. 

it's good  to know about it because newer shells like zsh have it enabled by default, and many build tools use it (this is how i found out about it)

vitalipe:
@dovron - great job! :)

12_glob_lab/globlab.txt
vitalipe: oh and as a side note I think that Ynon meant "in any folder *non-recursively* "... 
so ``` */*.c ``` would do the job.
 
but in the real world you'll want the recursive thing way more often :)

Issue #15, opened 2016-12-22
תיקיות שמתחילות באותיות גדולות


Aviyoosh:
echo [A-Z]* isn't displaying Upper cases, keep displaying lower cases.
what's the command?

ynonp:
The command is correct, but I guess your sort collation order is wrong (it happens). You can read more about it here:
http://teaching.idallen.org/net2003/06w/notes/character_sets.txt

The simple solution is in your shell to write `export LC_ALL=POSIX` or `export LC_ALL=C` (both will work). Later in the course we'll learn how to change this permanently

vitalipe:
It happens due to order of characters in a your locale.. [A-Z] means "match anything in that range", but you locale has the following range => ```aAbBcCdDeE........xXyYzZ``` 

note that the file named ```a.text``` will not be matched! because a comes before A :) 
to fix it you can either use what Ynon suggested or this little trick:
```LC_ALL=C bash -c 'echo [A-Z]*'```

you can also "alias" it, to the name for your choice like so:
```alias echo_all_upper="LC_ALL=C bash -c 'echo [A-Z]*'"```
then simply run:
```echo_all_upper```

alias is actually very common in bash, and you can inspect it with the builtin ```type``` command:
```
vitali@Tux:~/$ type ls
ls is aliased to `ls --color=auto'
vitali@Tux:~/$ type echo_all_upper 
echo_all_upper is aliased to `LC_ALL=C bash -c 'echo [A-Z]*''
``` 

let me know if that makes sense to you :)

vitalipe:
after reading @maximus777666's solution I realized that I'm getting old and  completely forgot about [[:upper:]]... and this should just work: ```ls [[:upper:]]``` :smile: 

there are also ```[:lower] [:space:] [:digit:]....``` and more...
to get the complete list just type: ``` man 7 glob```

PR #22, opened 2016-12-24

nmilgalter:


12_glob_lab/globlab.txt
```diff
@@ -4,26 +4,29 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
 1. Type in a glob pattern to match:
 
   1. All files whose name starts with a lowercase letter
-  Answer: 
+  Answer: [a-z]*.*
```
vitalipe, line 7: files don't need to have an extension... this will match all files that start with lower case and contain a dot in their name, not all files..

12_glob_lab/globlab.txt
```diff
@@ -4,26 +4,29 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 
   2. All files whose name starts with an uppercase letter
-  Answer:
+  Answer: [A-Z]*.*
```
vitalipe, line 10: same as above...

12_glob_lab/globlab.txt
```diff
@@ -4,26 +4,29 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
   6. All files with extension ".c" under any folder
-  Answer:
+  Answer: */*.c
 
 2. Write a command that would delete all files ending with a ".png" under any subdirectory.
+$ rm */*.png
 
 3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
 
+
```
vitalipe, line 29: you probably got *all* files and not just the ones that start with an uppercase - [here I try to explain why](https://github.com/tocodeil/unix1-course-examples/issues/15) tet me know in the issue if it makes sense to you..

vitalipe:
@nmilgalter I've added some comments about the usage of "*", let me know if you have any questions.. otherwise good job!

PR #25, opened 2016-12-24
ToCode Solution: תרגיל קבצים ותיקיות

everybodyls:


10_files_lab/fileslab.txt
```diff
@@ -8,30 +8,49 @@ Online: https://www.tocode.co.il/bundles/unix1/lessons/fileslab
   - .highscore
 
 Write here the commands you used:
+	sudo mkdir lab
```
vitalipe, line 11: I don't think you need root here.. ```sudo``` is a dangerous command and should be used only when needed.

as a general rule you should never use ```sudo``` when copying files in your home directory, and you should only use it when you get a permission denied error, and you understand the command that you're running :)

10_files_lab/fileslab.txt
```diff
@@ -8,30 +8,49 @@ Online: https://www.tocode.co.il/bundles/unix1/lessons/fileslab
...
 
 Write here the commands you used:
+	cd lab
+	sudo mkdir Music Misc Drivers
 
 3. Copy the file "/etc/passwd" to each subdirectory created in (2).
 
 Write here the commands you used:
 
+	sudo cp /etc/passwd /home/lab/Music
```
vitalipe, line 28: that's correct (except of the sudo which I mentioned above :P), but you can also use the shorthand for your home dir "~"

``` cp /etc/passwd ~/Music ```

(whatever you prefer, it's a matter of style)

vitalipe:
@caspib your solution was great except for the use of ```sudo``` 

please care with this command in the future, it may cause real damage to a 
running system (and your files). 

if you have any questions feel free to open an issue or PM me :)
(I'm closing this, but you can still comment)

PR #26, opened 2016-12-24

everybodyls:


07_cmd_lab/cmdlab.txt
```diff
@@ -2,12 +2,31 @@ Intro To Command Line Lab.
...
+       ·   numerical group ID
+
+       ·   user name or comment field
+
+       ·   user home directory
+
+       ·   optional user command interpreter
 
 3. In Unix, the /etc folder holds configuration files. Display all files in "/etc" folder ordered by file size (Hint: search in "man ls")
+	ls -l -S /etc
```
vitalipe, line 29: correct! :) 
also note that you can combine flags like so: ``` ls -Sl ```

07_cmd_lab/cmdlab.txt
```diff
@@ -2,12 +2,31 @@ Intro To Command Line Lab.
...
 
 4. Print current date/time formatted as: "Sun 21/03/2010, 14:30". Hint: read "man date" and search for the correct format string.
+	date -R
```
vitalipe, line 32: this is actually a pretty neat solution!  the idea was to use a format string here.. but well.. you outsmarted us ;)

vitalipe:
@caspib well done! :)

PR #27, opened 2016-12-24

everybodyls:


12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,25 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
-
+	ls *.c
 2. Write a command that would delete all files ending with a ".png" under any subdirectory.
-
+	rm *.png -r
 3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
-
-4. Search in "man ls" for an option to show only directory data so the command "ls _____ D*" should print only the word Doc
+	list of all files that contain Any single upper-case ascii alphabet letter, and all the rest of files
+4. Search in "man ls" for an option to show only directory data so the command "ls -D*/" should print only the word Doc
```
vitalipe, line 28: the flag ```-D``` is not the same as ```-d```.

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,25 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
+	ls [a-z]*
   2. All files whose name starts with an uppercase letter
   Answer:
-
+	ls [A-Z]*
   3. All files whose name ends in an "a", "e", "i", "o" or "u"
   Answer:
-
+	ls *[aeio] [u]
   4. All files whose name DOES NOT end in an "a", "e", "i", "o" or "u"
```
vitalipe, line 15: almost! I think you misread the question "u" should be inside the first [] :)

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,25 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
-
+	ls *[!aeio] [!u]
   5. All files with extension ".c" that are in a folder named "src"
   Answer:
-
+	ls /src *.c
   6. All files with extension ".c" under any folder
   Answer:
-
+	ls *.c
```
vitalipe, line 23: this will list all files in current directory try ```*/*.c``` (but note that it's not recursive - i,e won't work for sub directories)

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,25 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 2. Write a command that would delete all files ending with a ".png" under any subdirectory.
-
+	rm *.png -r
```
vitalipe, line 25: the -r flag is redundant here unless you want to delete directories that end with ".png".. note that ```mydir.png``` is actually a valid directory name... :)  

also it's a good practice to put your flags before the 
arguments ```rm -r *.png``` somewhat because of style and somewhat because  *.png glob is expended by the shell before the command is executed and if ever gonna write/debug shell scripts, it's gonna be hard to the -r see...

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,25 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
-
-4. Search in "man ls" for an option to show only directory data so the command "ls _____ D*" should print only the word Doc
+	list of all files that contain Any single upper-case ascii alphabet letter, and all the rest of files
```
vitalipe, line 27: all files that begin with *any* letter except of lower case "a" or all files that begin will an upper-case letter depends on your locale, [I try to explain here](https://github.com/tocodeil/unix1-course-examples/issues/15)

vitalipe:
@caspib I've added some notes to hopefully clear things up. Let me know if something is unclear or anything else really ;)

PR #32, opened 2016-12-25

MedinaVili:


12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,26 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
-
+echo [A-Z]*
   3. All files whose name ends in an "a", "e", "i", "o" or "u"
   Answer:
-
+echo *[a,e,i,o,u]
   4. All files whose name DOES NOT end in an "a", "e", "i", "o" or "u"
   Answer:
-
+echo *[!a,!e,!i,!o,!u]
```
vitalipe, line 17: almost! you only need one "!" at the beginning: 
 ```*[!aeiou] # , is optional ```

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,26 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
   5. All files with extension ".c" that are in a folder named "src"
   Answer:
-
+cd src
+echo *.c
   6. All files with extension ".c" under any folder
   Answer:
-
+find . -type f -name *.c
```
vitalipe, line 24: that's probably the best way to achieve it! :) 
but note that you're missing quotes on *.c, it should be '*.c' or "*.c".

this is due to the fact that the shell will expend the glob **before** it invokes ```find``` the same way it does with ```echo```

when you use quotes the shell will treat is as a simple string, and pass it to the find command.

you can verify it if you create a sub-directory and put a .c file inside.

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,26 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 2. Write a command that would delete all files ending with a ".png" under any subdirectory.
-
+find . -type f -name *.png -delete
```
vitalipe, line 26: same as above :)

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,26 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
-
-4. Search in "man ls" for an option to show only directory data so the command "ls _____ D*" should print only the word Doc
+there is invesible files so with ls-1 we cant see tham inspite of wildcards file.
```
vitalipe, line 28: all files that begin with *any* letter except of lower case "a" or all files that begin will an upper-case letter depends on your locale, [I try to explain here](https://github.com/tocodeil/unix1-course-examples/issues/15)  :) 

note that you can display hidden files with ```ls -l``` if you provide their name as a parameter, for example ```ls -l .hidden-file```

vitalipe:
@MedinaVili please refer to my comments about using quotes, otherwise good job! :)

PR #34, opened 2016-12-25

leotorjeman:


12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,34 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
-
+ls [a-z]*
   2. All files whose name starts with an uppercase letter
   Answer:
-
+ls [A-Z]*
   3. All files whose name ends in an "a", "e", "i", "o" or "u"
   Answer:
 
+  leo@leo-VirtualBox:~/lab$ ls *.[a,e,i,o,u]
```
vitalipe, line 15: when we say "files" we usually mean files & directories... this is because file **are** directories :) actually the zen of  unix is that (almost) everything has a file interface, even devices! (check /dev if you don't believe me ;) )   
 
also note that extensions in Unix are **completely** optional... for example ```/etc/passwd``` is a text file. also this command  won't really work..  you need to do this: ```*.*[aeiou]``` in order to find files (and directories!) that contain dot in their name..

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,34 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
+
   4. All files whose name DOES NOT end in an "a", "e", "i", "o" or "u"
   Answer:
+leo@leo-VirtualBox:~/lab$ ls *.[!a,e,i,o,u]
```
vitalipe, line 19: same as above.. :)

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,34 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 
   5. All files with extension ".c" that are in a folder named "src"
   Answer:
-
+ls *.c src/
```
vitalipe, line 23: here you list all the files in you current dir **and** all the files in ```src/``` your glob pattern is missing the files in src/, it should be ```src/*.js```

why this is important? well because shell globs are expended **before** the command is invoked by the shell, and here ls happens to list all files in a give directory, but it's not always the case.. 

consider this:
```
vitali@Tux:~/work/react-hut$ echo "echo is a builtin command that prints it's arguments"
echo is a builtin command that prints it's arguments

vitali@Tux:~/work/react-hut$ echo src
src

vitali@Tux:~/work/react-hut$ echo src/*.js
src/class-lists.js src/hut.js
```   
see that in the 3rd call to ```echo``` it printed file names? that's because, echo never saw "src.*.js"

vitalipe, line 23: here you list all the files in you current dir **and** all the files in ```src/``` your glob pattern is missing the files in src/, it should be ```src/*.js```

why this is important? well because shell globs are expended **before** the command is invoked by the shell, and here ls happens to list all files in a give directory, but it's not always the case.. 

consider this:
```
vitali@Tux:~/work/react-hut$ echo "echo is a builtin command that prints it's arguments"
echo is a builtin command that prints it's arguments

vitali@Tux:~/work/react-hut$ echo src
src

vitali@Tux:~/work/react-hut$ echo src/*.js
src/class-lists.js src/hut.js
```   
see that in the 3rd call to ```echo``` it printed file names? that's because, echo never saw "src.*.js"

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,34 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
   6. All files with extension ".c" under any folder
   Answer:
-
+find . -type f -name *.c
```
vitalipe, line 26: almost! after reading my comment above let's look at ```*.c``` find will never see ```*.c``` here try this:
```
vitali@Tux:~/work/react-hut$ echo -type f -name *.c
-type f -name *.c

vitali@Tux:~/work/react-hut$ touch crap-I-was-matched-by-a-glob.c

vitali@Tux:~/work/react-hut$ echo -type f -name *.c
-type f -name crap-I-was-matched-by-a-glob.c

vitali@Tux:~/work/react-hut$ echo -type f -name "*.c"
-type f -name *.c

```

yeah I know... very confusing... the golden rule is add quotes if you want the command to see it :)

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,34 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 2. Write a command that would delete all files ending with a ".png" under any subdirectory.
 
+find . -type f -name *.png -delete
+
```
vitalipe, line 30: same as above :)

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,34 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
 
+1 hidden files
```
vitalipe, line 33: all files that begin with *any* letter except of lower case "a" or all files that begin will an upper-case letter depends on your locale, [I try to explain here](https://github.com/tocodeil/unix1-course-examples/issues/15)

12_glob_lab/globlab.txt
vitalipe: btw just to clarify - the glob ```*.[a,e,i,o,u]``` will match any file that ends with ".a" ".e" ".i" ".o" ".u"...

vitalipe:
@leotorjeman  - I suggest that you watch the chapter again, It's a very confusing topic, I know! hopefully my comments cleared a few things... always feel free to open issues if you have any questions or email me directly -> vitali.pe@gmail.com

PR #36, opened 2016-12-25
ToCode Solution: תרגיל Links

MedinaVili:


18_links_lab/linkslab.txt
```diff
@@ -8,13 +8,15 @@ Preparation: Perform the following  (no need to write the commands you used):
...
   - Create a hard link named file4 that points to "file2"
   Answer: 
-
+ln file2 file4
   - Create a new directory named "tmp" in your home folder:
   Answer:
 
 2. Move "file3" to the new "~/tmp" directory you just created. Can you still use it to see the original "file1" ? Why not? How would you create a symbolic link that would survive the transfer to ~/tmp ?
-
+no, we changed file3 place so the link will not work anymore. we can use "ls -l" which survive thee transfer.
```
ynonp-test, line 20: Indeed a "hard" link created just with `ln` (without the -s) would survive the transfer, but counts as a hard link.

For a symbolic link to survive the move, it has to point to an absolute path. For example the following code creates a symbolic link that ALWAYS points to /etc/shells, even if moved:

```
ln -s /etc/shells my-link
```

So one option is to use the absolute path of the file for example:

```
ln -s /home/user/links/file1 file3
```

But, if you're already in the directory /home/user/links (because you cd to it earlier), you can use the even shorter version:

```
ln -s $(realpath file1) file3
```

Its meaning: Run the command `realpath file1`, take its output and use it as the argument. Since realpath returns the absolute path to a file it yields the same result as the longer version.

18_links_lab/linkslab.txt
```diff
@@ -8,13 +8,15 @@ Preparation: Perform the following  (no need to write the commands you used):
...
-
+ln file2 file4
   - Create a new directory named "tmp" in your home folder:
   Answer:
 
 2. Move "file3" to the new "~/tmp" directory you just created. Can you still use it to see the original "file1" ? Why not? How would you create a symbolic link that would survive the transfer to ~/tmp ?
-
+no, we changed file3 place so the link will not work anymore. we can use "ls -l" which survive thee transfer. 
 3. How can you tell that "file2" and "file4" refer to the same file?
+ls -l [option to sort] and then we can the last date of Modifying and also how names there are to Specific file.
```
ynonp, line 22: Better to use `ls -li` - as that prints also the special "inode index" of the file. A unique number that equals only for hard links

PR #42, opened 2016-12-26

Meiron:


12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,90 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
 
   1. All files whose name starts with a lowercase letter
   Answer: 
+  #amir@ubuntu:~/wildcards$ find ./ -name "[[:lower:]]*.*"
```
ynonp, line 8: No need for `./` a simple `find . -name ...` is enough

ynonp: Also the task was to find files that start with a lowercase letter. Not all files in unix have `.` (period) sign in them, for example file `/etc/shells`, `/bin/ls` etc. have no extension. 
It's a normal thing on Unix systems not to have extension for every file.

Searching for `[[:lower:]]*.*` only searches for files with extensions, but on Unix we prefer the `[[:lower:]]*` to search for all files starting with a lowercase letter, with or without file extensions

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,90 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
+  #./src
+  #./src/player.c
+  #./src/main.c
+  #./include
+  #./include/player.h
+
 
   5. All files with extension ".c" that are in a folder named "src"
   Answer:
+  #amir@ubuntu:~/wildcards$ find ./src  -name "*.[c]"
```
ynonp, line 49: The `[c]` is not required here, since c is the only option. Better to just use `*.c`

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,90 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
   Answer:
+  #  amir@ubuntu:~/wildcards$ find ./  -name "*.[c]"
+  #./Doc/test.c
+  #./lib/zlib.c
+  #./src/player.c
+  #./src/main.c
+
 
 2. Write a command that would delete all files ending with a ".png" under any subdirectory.
+  #find ./  -name "*.png" -exec rm -f {} \;
```
ynonp, line 64: Correct! Though I'd use "rm -i" instead so you get a prompt before deleting each file and so can be sure you're not deleting things you need

12_glob_lab/globlab.txt
```diff
@@ -5,25 +5,90 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
+ #-rw-rw-r-- 1 amir amir 0 Dec 25 09:50 zlib.c
 
-3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
+ #src:
+ #total 0
+ #-rw-rw-r-- 1 amir amir 0 Dec 25 09:50 main.c
+ #-rw-rw-r-- 1 amir amir 0 Dec 25 09:50 player.c
 
 4. Search in "man ls" for an option to show only directory data so the command "ls _____ D*" should print only the word Doc
+ #amir@ubuntu:~/wildcards$ ls -a | grep [D*]
```
ynonp, line 92: This works here but not in the general case. A better solution is to use `-d` which tells ls not to show content of a directory. Full command `ls -l D*`

ynonp (approved):

PR #48, opened 2016-12-28

shlomitTal:


07_cmd_lab/cmdlab.txt
```diff
@@ -3,11 +3,34 @@ Questions online: https://www.tocode.co.il/bundles/unix1/lessons/cmdlab
...
 
-2. View the file /etc/passwd on your machine with the command "cat /etc/passwd". Then read about the file in "man 5 passwd". What information is held in the file? What are the fields in each line?
+ 
 
 3. In Unix, the /etc folder holds configuration files. Display all files in "/etc" folder ordered by file size (Hint: search in "man ls")
+answer: ls -S /etc
 
 4. Print current date/time formatted as: "Sun 21/03/2010, 14:30". Hint: read "man date" and search for the correct format string.
+answer:
+the commands are:
```
vitalipe, line 33: all you need is ```date "+%a %d/%m/%Y, %H:%M"```.

first you don't need a variable here, try this:
``` echo $(date +"%a %d/%m/%Y, %H:%M")```

also note that  $() will run a command "capture" it's output and assign it to the result of the expression... (you should learn about it in the  chapter about shell globs..)

it's useful when you want to assign the result of a command as an argument for another command, for example:
``` 
vitali@Tux:~$ which rm
/bin/rm

vitali@Tux:~$ ls -lh $(which rm)
-rwxr-xr-x 1 root root 59K פבר 18  2016 /bin/rm
```

the ```echo``` command will just print it's parameters, this is useful for scripts:
```
vitali@Tux:~$ echo "hey!"
hey!
```
 when used alone, they cancel each other... :)

vitalipe:
@shlomitTal please take note of my comment about echo and $(), and if you have any questions feel free to open an issue or PM me.. otherwise good job :)

PR #51, opened 2016-12-28

Aviyoosh:


12_glob_lab/globlab.txt
```diff
@@ -3,27 +3,36 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
+1. All files whose name starts with a lowercase letter
   Answer: 
 
-  2. All files whose name starts with an uppercase letter
+  echo [a-z]*
+2. All files whose name starts with an uppercase letter
   Answer:
 
-  3. All files whose name ends in an "a", "e", "i", "o" or "u"
+  LC_ALL=C bash -c 'echo [A-Z]*'
```
ynonp, line 14: Cool. I'd use `export LC_ALL=C` before doing the exercise which affects all commands in a session

12_glob_lab/globlab.txt
```diff
@@ -3,27 +3,36 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
   Answer:
 
-  5. All files with extension ".c" that are in a folder named "src"
+  echo *[!aeiou]
+5. All files with extension ".c" that are in a folder named "src"
   Answer:
 
-  6. All files with extension ".c" under any folder
-  Answer:
+ ls *.c src/
```
ynonp, line 26: The command you typed shows all *.c files AND files inside src. You were asked to combine so this is a better solution: `ls src/*.c`

12_glob_lab/globlab.txt
```diff
@@ -3,27 +3,36 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
-  5. All files with extension ".c" that are in a folder named "src"
+  echo *[!aeiou]
+5. All files with extension ".c" that are in a folder named "src"
   Answer:
 
-  6. All files with extension ".c" under any folder
-  Answer:
+ ls *.c src/ 
+6. All files with extension ".c" under any folder
+  Answer: ?:(
```
ynonp, line 28: Seeing the previous answer, it's cool that you can improve it by replacing src with * to get any folder, so the following works: `ls */*.c`

12_glob_lab/globlab.txt
```diff
@@ -3,27 +3,36 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
+ 
 
 2. Write a command that would delete all files ending with a ".png" under any subdirectory.
 
-3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
+ Amswer: find . -type f -name *.png -delete
+3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
+
 
 4. Search in "man ls" for an option to show only directory data so the command "ls _____ D*" should print only the word Doc
```
ynonp, line 38: The switch is -d: Normally ls shows all files inside a directory when given, so `ls Doc` would not show Doc but a list of all files inside. With wildcards this behavior is a bit confusing, so we have -d that tells ls to show only folder details and not content. `ls -d D*` will just print Doc, and `ls -ld D*` will show the long list with Doc permissions and details

ynonp (approved):

PR #53, opened 2016-12-29
ToCode Solution: תרגיל Links

Aviyoosh:


18_links_lab/linkslab.txt
```diff
@@ -2,19 +2,32 @@ Preparation: Perform the following  (no need to write the commands you used):
...
 
-  - Create a new directory named "tmp" in your home folder:
-  Answer:
+  ln file2 file4
+- Create a new directory named "tmp" in your home folder:
+  
+Answer:
 
-2. Move "file3" to the new "~/tmp" directory you just created. Can you still use it to see the original "file1" ? Why not? How would you create a symbolic link that would survive the transfer to ~/tmp ?
+   mkdir /home/avia/tmp
```
ynonp, line 25: Correct, and also worth knowing about ~ which is a shorthand to home directory, so this one works too: `mkdir ~/tmp`

18_links_lab/linkslab.txt
```diff
@@ -2,19 +2,32 @@ Preparation: Perform the following  (no need to write the commands you used):
...
+
+2. Move "file3" to the new "~/tmp" directory you just created. Answer: mv home/avia/lab4/links/file3 /home/avia/tmp
+Can you still use it to see the original "file1" ? No. Why not? Because when moving synbolic link to different directory - the link will not work for the original file anymore.
+How would you create a symbolic link that would survive the transfer to ~/tmp ?
 
-3. How can you tell that "file2" and "file4" refer to the same file?
+ln -s /home/avia/lab4/links/file1 /home/avia/tmp/file4
+
+3. How can you tell that "file2" and "file4" refer to the same file? ls -l gives us output for the files that exist in the folder, which we can see that file 2 and file 4 have the same date+permissions+ 2 = numbers of names to the same file
```
ynonp, line 33: The magic switch is `-i` - it shows a special unique index of a file that is only equal if the files are hard links to one another. Use `ls -li file1 file2 file3 file4` and watch the inode number, you'll see files 2 and 4 have the same

18_links_lab/linkslab.txt
```diff
@@ -2,19 +2,32 @@ Preparation: Perform the following  (no need to write the commands you used):
...
-  Answer:
+  ln file2 file4
+- Create a new directory named "tmp" in your home folder:
+  
+Answer:
 
-2. Move "file3" to the new "~/tmp" directory you just created. Can you still use it to see the original "file1" ? Why not? How would you create a symbolic link that would survive the transfer to ~/tmp ?
+   mkdir /home/avia/tmp
+
+2. Move "file3" to the new "~/tmp" directory you just created. Answer: mv home/avia/lab4/links/file3 /home/avia/tmp
```
ynonp, line 27: For a symbolic link to survive the move, it has to point to an absolute path. For example the following code creates a symbolic link that ALWAYS points to /etc/shells, even if moved:

```
ln -s /etc/shells my-link 
```

So one option is to use the absolute path of the file for example:

```
ln -s /home/user/links/file1 file3 
```

But, if you're already in the directory /home/user/links (because you cd to it earlier), you can use the even shorter version:
```
ln -s $(realpath file1) file3 
```

Its meaning: Run the command realpath file1, take its output and use it as the argument. Since realpath returns the absolute path to a file it yields the same result as the longer version.

ynonp (approved):

PR #54, opened 2016-12-30
ToCode Solution: תרגול הרשאות

yanivmix:


15_perms_lab/permslab.txt
```diff
@@ -1,15 +1,17 @@
...
 
   - change permission of file src/main.c to r--r--r-- using chmod letter notation
-
+chmod a=r src/main.c
   - change permission of file include/player.h to r-xrw--r- using chmod octal (numbers) notation
-
+chmod 564 src/player.c
   - change permission of all files in Doc folder (without changing the permissions of Doc folder itself) to rw-------
-
+chmod 777 Doc/*.*
```
ynonp, line 8: No need for the `*.*` - In unix not all files have extension so we use the simpler `Doc/*`

Also you were asked to grant permission rw only, so 777 is the wrong number here. You'll need: `chmod 600 Doc/*`

15_perms_lab/permslab.txt
```diff
@@ -1,15 +1,17 @@
...
   - change permission of "lib" folder and all files and subdirectories inside it to rwxrwxrwx
-
+chmod 777 l*/*.* or chmod -R 777 lib
 2. create a new directory named "test" and inside an empty file named "secret". Change permissions of the file "secret" to ---------. Can you delete the file secret? can you delete the folder test? why?
-
+I can delete the empty file: rm test/secret
+Cannot delete test, since it is a folder (Need to use rm -r)
```
ynonp, line 13: Not exactly. It is a folder and you do need `rm -r` but that's not enough. As `test` folder has no write permission, you can't delete the file `secret` that lies inside, and without it being deleted, the folder itself can't be deleted.

15_perms_lab/permslab.txt
```diff
@@ -1,15 +1,17 @@
...
 3. Read the output of "help umask". Explain in your own words what umask does, and how would you use it to set new files permissions to ------r--
-
+Gives you the option to change the defaults riths upon creation of files and folders
+There is a need to change the umask to 003, in order to set the files permissions to ------r--
```
ynonp, line 16: umask sets a negative bit mask, so we'll want the opposite. An enabled bit in the umask is actually a `-`, i.e. no permission on the result. So the correct umask here is:
`umask 662`
Turning off rw for owner and group, and write for other. What's left is just the "r" permission for others

ynonp (approved):

ynonp:
I just noticed there was a mistake in the question regarding the directory with write protected file inside. The question should have asked you to remove permissions from `test` folder itself (not the file inside). As it goes please ignore the question and related comment. 

Sorry for the confusion.

PR #55, opened 2016-12-30
ToCode Solution: תרגול הרשאות

dovron:


15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,61 @@
...
+tocode@tocode-VirtualBox:~/wildcards$ ll lib/
+total 8
+drwxrwxrwx 2 tocode tocode 4096 Dec 24 14:18 ./
+drwxrwxr-x 7 tocode tocode 4096 Dec 24 14:19 ../
+-rwxrwxrwx 1 tocode tocode    0 Dec 24 14:18 zlib.c*
+  
 2. create a new directory named "test" and inside an empty file named "secret". Change permissions of the file "secret" to ---------. Can you delete the file secret? can you delete the folder test? why?
 
-3. Read the output of "help umask". Explain in your own words what umask does, and how would you use it to set new files permissions to ------r--
+Yes, it possible because we have admin permissions
```
ynonp, line 53: Not exactly. Using normal "rm -r test" fails no matter which user you are because the file "secret" inside prevents the folder from being deleted. As there are no write permissions on the folder, rm -r can't automatically remove all files inside.

But as you said it is possible to delete if we first grant write+execute permission on the folder itself using `chmod 777 test`. Only then can we `rm -rf test` successfully

15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,61 @@
...
 
+3. Read the output of "help umask". Explain in your own words what umask does, and how would you use it to set new files permissions to ------r--
 
+umask 773
+tocode@tocode-VirtualBox:~/wildcards$ umask -S
+u=,g=,o=r
+tocode@tocode-VirtualBox:~/wildcards$ touch test
+tocode@tocode-VirtualBox:~/wildcards$ ll test
+-------r-- 1 tocode tocode 0 Dec 30 15:34 test
```
ynonp, line 62: 👍

ynonp (approved):

ynonp:
I just noticed there was a mistake in the question regarding the directory with write protected file inside. The question should have asked you to remove permissions from `test` folder itself (not the file inside). As it goes please ignore the question and related comment. 

Sorry for the confusion.

PR #57, opened 2016-12-30
ToCode Solution: תרגיל Links

dovron:


18_links_lab/linkslab.txt
```diff
@@ -4,17 +4,89 @@ Preparation: Perform the following  (no need to write the commands you used):
...
+python
+java
+shell script
+
+yes, we can see file4 - becuse of hard links.
+file2 is still ok , without any error
+
 3. How can you tell that "file2" and "file4" refer to the same file?
+
+they have the same ID (1)
```
ynonp, line 87: That column is not an ID but the total number of hard links to a file. The fact that "1" is written there is a bit alarming, since if hard links were created correctly it should have been "2".

Scrolling up I can see you created the hard link with `ln file1 file4` which probably explains it (should have been `ln file2 file4`).

The command `ls -li` prints a unique ID for each file called inode number. Try to use that and you'll see the numbers on the first column such as:

```
150619833 -rw-r--r--  1 ynonperek  staff  0 Jan  3 11:45 fil1
150619834 -rw-r--r--  2 ynonperek  staff  0 Jan  3 11:45 file2
150619834 -rw-r--r--  2 ynonperek  staff  0 Jan  3 11:45 file4
```

As you can see file2 and file4 have the same inode number

ynonp (approved):

PR #59, opened 2016-12-31

yanivmix:


22_pipeslab/pipeslab.txt
```diff
@@ -1,21 +1,23 @@
...
 
   4. Modify (1) and (2) to count files only (no directories)
   Answer:
+1: ls -l | tail -n +2 | cut -c1 | grep - | cut -c4 | grep x | wc | cut -c6-7
+2: ls -l | tail -n +2 | cut -c1 | grep - | cut -c4,7,10 | grep x | wc | cut -c6-7
 
-  5. The command /sbin/ifconfig shows information regarding the machine's network interfaces. Use the command to print only the IP address of the machine. Remember not to show the special loopback address 127.0.0.1
-
+5. The command /sbin/ifconfig shows information regarding the machine's network interfaces. Use the command to print only the IP address of the machine. Remember not to show the special loopback address 127.0.0.1
+/sbin/ifconfig | grep 'inet ' | head -1
```
ynonp, line 18: I suggest adding `grep -v 127.0.0.1` to the pipeline to remove 127.0.0.1 from the result. Also a `grep -v inet6` is sometimes required to remove IPv6 addressed. In sum you'd want:

```
/sbin/ifconfig | grep inet | grep -v inet6 | grep -v 127.0.0.1 | head -1
```

22_pipeslab/pipeslab.txt
```diff
@@ -1,21 +1,23 @@
...
 6. Joe wanted to sort a text file in place so he tried to run the command "sort myfile.txt > myfile.txt". What happened to the file? Why?
-
+all the values that were in the file were deleted. Problem is that he tries to run sort on a file and not in the content. Should be something like "cat myfile.txt | sort"
```
ynonp, line 20: Not exactly. `sort file` works very well and prints its contents sorted. Also you can try `cat file.txt | sort > file.txt` and get the file deleted again.

The problem is the order bash performs output redirects. A `>` is first executed and truncates contents of the file. Then everything else is read and executed, but unfortunately file is already empty so cat / sort see no contents.

Sort has a switch for such a use case: `-o` Using it you can sort a file in place: `sort file.txt -o file.txt`

22_pipeslab/pipeslab.txt
```diff
@@ -30,5 +32,9 @@ $ who | cut -d ' ' -f 3 | sort | uniq -c
 What was printed? Why? Fix the command line to get the output:
 2 Dec
 2 Nov
+Only the result of the uniq number from the count operation.
+It was because of the parameter "f" that was sent to the sort and uniq
+who | cut -d ' ' -f 17 | sort | uniq -c
+
```
ynonp, line 38: Partially works. `who` has different output style on each machine and the number 17 which works for you may not be correct for someone else.

The reason cut is wrong is its unable to work well with duplicate spaces. A better solution would be to remove the duplicate spaces before the command and use:

` who | tr -s ' ' | cut -d ' ' -f 3 | sort | uniq -c`

ynonp (approved):

PR #60, opened 2016-12-31
ToCode Solution: תרגול find

yanivmix:


24_find_lab/findlab.txt
```diff
@@ -1,24 +1,25 @@
...
+find ./tmp -user ym
   -. All files and directories not owned by you in /tmp folder
-
+find ./tmp ! -user ym
   -. All files, folders and links in your home folder whose name ends with a ".c"
-
+find . -name "*.c"
   -. All files in your home folder whose name starts with an "a", "e", "i", "o" or "u"
-
+find . -name "a*.*" -o -name "e*.*" -o -name "i*.*" -o -name "o*.*" -o -name "u*.*"
```
ynonp, line 10: 1. Don't use `*.*` in Unix as a default. Many files don't have extensions so a simple `*` is better. 

2. A nicer glob is the bracket, so you can use `find . -name "[aeiou]*"`

ynonp (approved):

PR #66, opened 2017-01-02

everybodyls:


30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,19 @@
 Use grep to print only the lines from "input" for which:
 
 1. There is a binary number in the line
+	cat input |grep -a [0-9]
```
ynonp, line 4: Not sure `-a` is required here (usually it's used for grepping binary files)

ynonp: A binary number is one that has 0s and 1s only. For example the line `The number 01010100 is binary` has a binary number while the line `The number 90 is not` does not have one.

A regular expression would need to check if a single word in the line is completely made up from 0s and 1s only. The grep command for that is: `grep -w [01]+`
Its meaning: Search for a sequence (one or more) of the characters 0 and 1, that appear as a word in itself, due to the -w switch.

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,19 @@
...
 
 2. There is an even binary number in the line
-
+	cat input |grep -a *[0-9]
```
ynonp, line 7: An even binary number is a binary number that ends in 0. Given the answer of (1), the correct solution here is `grep -w [01]*0`. Its meaning:

Search for a sequence of 0 or more characters from group (0,1), followed by a single 0, and as its own word.

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,19 @@
...
 3. Line includes a hexadecimal number
-
+	cat input |grep '[xNN]'
```
ynonp, line 9: A hexa decimal number can either start in 0x or not, and in any case it is followed by normal digits. The regexp is therefore '(0x)?[0-9]+' which means the sequence 0x may appear (0 or 1 times), and after it a sequence of 1 or more normal digits

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,19 @@
...
 4. Line starts with a lowercase letter
-
+	cat input |grep ^[a-z]
 5. Line is not empty
-
+	cat input|grep -v -e '^$'
 6. Line starts with a lowercase letter and ends in a number
-
+	cat input |grep ^[a-z] $[0-9]
```
ynonp, line 15: A space in regular expression translates to a space, not to an AND. To create an AND you need to add another grep for example: `cat input | grep '^[a-z]'|grep '[0-9]$'`

Also, note $ must appear as the last character for it represents end-of-line

ynonp (approved):

PR #70, opened 2017-01-02

maximus777666:


22_pipeslab/pipeslab.txt
```diff
@@ -3,19 +3,106 @@ For each of the following, type a single command line that performs:
   1. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner.
   Answer: 
 
+maxim@maxim-virtual-machine:~$ ls -l | grep '^...x' | cut -d' ' -f3,9
```
vitalipe, line 6: what's clever! but you need to count the lines.. there's a nice command called ``wc`` (Word Count):
```
wc -l # count lines
wc -c # count characters
wc -w # count words
```

22_pipeslab/pipeslab.txt
```diff
@@ -3,19 +3,106 @@ For each of the following, type a single command line that performs:
...
+maxim Videos
+maxim wildcards
+
+
+  
   2. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner, group and others.
   Answer: 
 
+  maxim@maxim-virtual-machine:~$ ls -l | grep '^...x..x..x' | cut -d' ' -f3,9
+maxim Desktop
```
vitalipe, line 25: again ```wc``` :)

22_pipeslab/pipeslab.txt
```diff
@@ -3,19 +3,106 @@ For each of the following, type a single command line that performs:
...
+maxim Templates
+maxim tmp
+maxim Videos
+maxim wildcards
+
+  
   3. Print the number of lines from "ls -l" which represent paths that were modified today (no matter which day we run the command)
   Answer: 
+  
+maxim@maxim-virtual-machine:~$ ls -l | grep $(date | cut -d' ' -f3)
```
vitalipe, line 41: the usage of ```cut``` here is very problematic here.. there is just too much that can break(for example on my machine -f4 works) .. 

I would use the format  string to date that you learned in chapter 1:
```date "+%b %e```
and actually call date twice, like so:
```
grep "$(date +%b)\s* $(date +%e)"
```

the ```\s* ``` part means at least one space

and then ```wc``` ;)

File: 22_pipeslab/pipeslab.txt [hunk 4]
```diff
@@ -3,19 +3,106 @@ For each of the following, type a single command line that performs:
...
+-rwxrwxrwx  1 maxim maxim     0 Jan  2 20:24 file.test
+-rwxrwxr-x  1 maxim maxim   126 Dec 29 17:03 hw.7
+
+
+  
+  5. The command /sbin/ifconfig shows information regarding the machine's network interfaces. Use the command to print only the IP address of the machine. 
+  Remember not to show the special loopback address 127.0.0.1
+
+
+maxim@maxim-virtual-machine:~$ ifconfig | egrep '([0-9]{1,3}\.){3}[0-9]{1,3}' | grep Bcast | cut -d ':' -f 2 | cut -d' ' -f 1                                           10.61.40.39
```
vitalipe, line 94: that's pretty clever.. but I would not use the ```egrep``` pattern here maybe just ```grep "inet addr:"```

22_pipeslab/pipeslab.txt
```diff
@@ -3,19 +3,106 @@ For each of the following, type a single command line that performs:
...
+10.61.40.39
+
+  
 6. Joe wanted to sort a text file in place so he tried to run the command "sort myfile.txt > myfile.txt". What happened to the file? Why?
 
+Blank file after running the command
+second form running the command, filename doesn't delete anything from the file - it only creates an empty file if one did not exist,
```
vitalipe, line 101: bash will process the redirection first, then it executes the command...so by the time ```sort``` looks at the file, it's empty!

PR #74, opened 2017-01-03

everybodyls:


vitalipe:
@caspib well done! :)

vitalipe:
BTW not sure if Ynon mentioned it, but aliases are used all over the place in modern linux systems..  the command ```type``` allows you to see the alias:
```
vitali@Tux:~/tmp$ type ls
ls is aliased to `ls --color=auto'
vitali@Tux:~/tmp$ type grep
grep is aliased to `grep --color=auto'
vitali@Tux:~/tmp$ 
```

PR #75, opened 2017-01-04
ToCode Solution: תרגול find

maximus777666:


24_find_lab/findlab.txt
```diff
@@ -1,24 +1,140 @@
...
+drwx------ 3 root  root  4096 Jan  4 09:39 systemd-private-07794fef300641d3a64d3c789b0d3e70-systemd-timesyncd.service-JiwDox
+-rw-rw-r-- 1 maxim maxim    0 Jan  4 09:42 test1.maxim
+-rw-rw-r-- 1 maxim maxim    0 Jan  4 09:42 test2.maxim
+-rw-rw-r-- 1 maxim maxim    0 Jan  4 09:42 test.maxim
+
+
+
+  maxim@maxim-virtual-machine:/tmp$ find . -exec ls -l \; | grep $(whoami) | uniq
+  
+  ** I don't understabd why it print all 4 times, and why "uniq" is not working.
```
vitalipe, line 22: it will invoke ```ls -l``` each time a file is matched (and in this case for every file!)... 

as for ```uniq``` note that this command **by default** will remove identical and adjacent lines, here watch this:
```
vitali@Tux:~/tmp$ cat demo.txt 
wow!
wow!
wow!
moooo!
wow!
vitali@Tux:~/tmp$ uniq demo.txt 
wow!
moooo!
wow!
vitali@Tux:~/tmp$ 
```

you need to pass the ```-u``` flag, if you want unique lines:
```
vitali@Tux:~/tmp$ uniq -u demo.txt 
moooo!
wow!
```

vitalipe: also note that ```-exec``` option allows you to pass the matched file name to your command. just add ```{}``` and it will replace it for you. 
try: ```find ./ -exec echo {} \;``` to see it in action :) 

also your solution with grep & whoami is very creative, but ```find``` supports an option called ```-user``` that will do all the hard work ;)

vitalipe: and there's a ```-type``` flag that allows you specify file type :)
so a complete command might look like this:
``` find ./ -type f -user maxim```

File: 24_find_lab/findlab.txt [hunk 2]
```diff
@@ -1,24 +1,140 @@
...
+-rw-rw-r-- 1 maxim maxim     2 Jan  4 10:08 a.txt.bak
+-rw-rw-r-- 1 maxim maxim     2 Jan  4 10:02 b.txt
+-rw-rw-r-- 1 maxim maxim     2 Jan  4 10:08 b.txt.bak
+-rw-rw-r-- 1 maxim maxim     2 Jan  4 10:02 c.txt
+-rw-rw-r-- 1 maxim maxim     2 Jan  4 10:08 c.txt.bak
+-rw-rw-r-- 1 maxim maxim     0 Jan  4 09:42 test1.maxim
+-rw-rw-r-- 1 maxim maxim     0 Jan  4 09:42 test2.maxim
+-rwxrwxrwx 1 maxim maxim     0 Jan  4 09:42 test.maxim
+
+It prints all X4 times
```
vitalipe, line 37: ^ same as above ;)

24_find_lab/findlab.txt
```diff
@@ -1,24 +1,140 @@
...
+  
+  
 
   -. All files and directories not owned by you in /tmp folder
+  
+  user root,test
```
vitalipe, line 43: there's a cool neat feature to ```find``` that allows you to invert the next flag it's called ```-not``` so this command:
``` find ./ -not -user maxim``` 
will print all the files that are not owned by you.

File: 24_find_lab/findlab.txt [hunk 4]
```diff
@@ -1,24 +1,140 @@
...
+./wildcards/src/main.c
+./wildcards/src/player.c
+./wildcards/Doc/test.c
+./test1.c
+./lab/test2.c
+
 
   -. All files in your home folder whose name starts with an "a", "e", "i", "o" or "u"
+  
+  maxim@maxim-virtual-machine:~$ find . * | grep "^[a,e,i,o,u]"
```
vitalipe, line 66: it works.. but note that "*" is expended by the shell **before** find ever see it! that makes ```find``` redundant here.. also you can pass the same pattern to ```find -name``` just make sure it's quoted: ```find ./ -name "[aeiou]"```

24_find_lab/findlab.txt
```diff
@@ -1,24 +1,140 @@
...
+a.test
+e.test
+ipconfig
+o.test
+u.test
+
+    -. All files in your home directory whose size is above 10MB
+  
+  maxim@maxim-virtual-machine:~$ find . -size +10M
```
vitalipe, line 75: yes! that's an extremely useful command BTW

24_find_lab/findlab.txt
```diff
@@ -1,24 +1,140 @@
...
 
-2. Create a hard link to a file of your choice and put it in a subdirectory under your home folder. How can you use "find" to locate the hard link given the original file?
+  maxim@maxim-virtual-machine:~$ find . -mtime -30 -size +10M
+./il.archive.ubuntu.com_ubuntu_dists_xenial_universe_binary-amd64_Packages
+
+
+2. Create a hard link to a file of your choice and put it in a subdirectory under your home folder. 
+How can you use "find" to locate the hard link given the original file?
+
+maxim@maxim-virtual-machine:~$ find . -type f -links +1
```
vitalipe, line 96: that's partially the answer.. the question was referring to how to get from a given filename  to all it's hard links.. and of course ```find``` has a flag for that also! :P   it's called ```-samefile``` try it ;)

24_find_lab/findlab.txt
```diff
@@ -1,24 +1,140 @@
...
 What was the command?  
 
+maxim@maxim-virtual-machine:/tmp$ touch a.txt b.txt c.txt
+maxim@maxim-virtual-machine:/tmp$ echo a > a.txt && echo b > b.txt && echo c > c.txt
+maxim@maxim-virtual-machine:/tmp$ cat a.txt b.txt c.txt
+a
+b
+c
+
+maxim@maxim-virtual-machine:/tmp$ find . -name "*.txt" -exec cp {} "{}".bak \;
```
vitalipe, line 123: that's correct! but I would use a more specific pattern 
like ```?.txt``` or better yet ```[abc].txt```

vitalipe:
@maximus777666 Hopefully after reading my comments things are a bit more clear now. As always feel free to open issues or PM me if you have any questions :)

PR #78, opened 2017-01-05

dovron:


22_pipeslab/pipeslab.txt
```diff
@@ -3,19 +3,42 @@ For each of the following, type a single command line that performs:
   1. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner.
   Answer: 
 
+  tocode@tocode-VirtualBox:~$ ls -alF | grep tocode | cut -c4 | grep x | wc | cut -d' ' -f6
```
vitalipe, line 6: hmm.. 

```grep tocode``` is not needed here because we want "execute permission to **their** owner". not just for your user... 

```wc | cut -d' ' -f6``` using cut here is not needed because you can just ask ```wc``` to 
print the lines with ```wc -l```  also I kinda hate the -f option with cut because it's too flaky.. 
for example on my machine it works with ```-f7```. so I don't recommend  using cut  to often.

Comment (dovron) [reply 22_pipeslab/pipeslab.txt:6]: what do you mean "execute permission to their owner"?
how can I do it?

File: 22_pipeslab/pipeslab.txt [hunk 2]
```diff
@@ -3,19 +3,42 @@ For each of the following, type a single command line that performs:
...
+  20
+  grep x is not must
+  
   2. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner, group and others.
   Answer: 
-
+ 
+  ls -alF | grep tocode | grep x..x..x | wc | cut -d' ' -f6
```
vitalipe, line 13: Same issue with cut and wc.. the grep thing is actually very clever! 

Just a tiny note about ```grep```:
I recommend always using quotes with grep patterns, here it doesn't matter, but if you if pass in something that looks like a shell glob it will be expended before the command. And that's not something that you usually want.

22_pipeslab/pipeslab.txt
```diff
@@ -3,19 +3,42 @@ For each of the following, type a single command line that performs:
...
+  
+  
   3. Print the number of lines from "ls -l" which represent paths that were modified today (no matter which day we run the command)
   Answer: 
 
+  ls -alF | grep tocode | grep Jan..2 | wc | cut -d' ' -f7
```
vitalipe, line 19: The idea here is that the date should not be hard coded.. once we 
play enough with the date command, we can see that:
```
date "+%b" # will print the current month
date "+%e" # will print the day  (1-31), space padded
```
Now we just need to feed their input  into grep and count lines, let's use $():
```
echo  "$(date +%b) $(date +%e)" # will print "ינו  7" (without quotes of course..)
ls -l | grep "$(date +%b) $(date +%e)" | wc -l
```
Now as a last touch, the fact the we check for a single white space kinda bothers me, let's fix it:  
```ls -l | grep "$(date +%b)\s* $(date +%e)" | wc -l``` 
The pattern ```\s*``` means 0 or more spaces, and we follow it with a white-space.

There are many ways to solve this, this is just one of them :)

File: 22_pipeslab/pipeslab.txt [hunk 4]
```diff
@@ -3,19 +3,42 @@ For each of the following, type a single command line that performs:
...
+  
   4. Modify (1) and (2) to count files only (no directories)
   Answer:
+  
+	ls -alF | grep tocode | grep -v / | cut -c4 | grep x | wc | cut -d' ' -f7
+	
+	ls -alF | grep tocode | grep -v / | grep x..x..x | wc | cut -d' ' -f7
```
vitalipe, line 26: note the first character in ```ls -l``` when it's a directory and when it's a file...

22_pipeslab/pipeslab.txt
```diff
@@ -3,19 +3,42 @@ For each of the following, type a single command line that performs:
...
 
+	
   5. The command /sbin/ifconfig shows information regarding the machine's network interfaces. Use the command to print only the IP address of the machine. Remember not to show the special loopback address 127.0.0.1
+	
+	date:
+	ls -alF | grep "$(date '+%b %e')"
```
vitalipe, line 32: damn... that's the previous answer.. please keep them in order..

22_pipeslab/pipeslab.txt
```diff
@@ -3,19 +3,42 @@ For each of the following, type a single command line that performs:
...
+	
+	ifconfig | grep broadcast | tr -s ' ' | cut -d' ' -f3
```
vitalipe, line 34: I think it's ```Bcast``` ;)

22_pipeslab/pipeslab.txt
```diff
@@ -3,19 +3,42 @@ For each of the following, type a single command line that performs:
...
 
+  
 6. Joe wanted to sort a text file in place so he tried to run the command "sort myfile.txt > myfile.txt". What happened to the file? Why?
 
+	it will overwrite the existing file and it will earsed.
+		in order to prevent it the command: set -o noclobber should be used.
```
vitalipe, line 40: neat! I didn't knew that this option existed! :D

22_pipeslab/pipeslab.txt
vitalipe: It simply means that the user that owns the file can execute it.. remember that there are 3 
sets of permissions? owner/group/other ? here we want to count the files that have 
execute permission in owner :)

vitalipe: btw I didn't see any email or notification for this message this is weird...

22_pipeslab/pipeslab.txt
dovron: you didnt watch the movie? :)

22_pipeslab/pipeslab.txt
dovron: I think you have issue with the order in the web (in hebrew) and in the files...

22_pipeslab/pipeslab.txt
dovron: why Bcast?

enp0s3: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 10.0.2.15  netmask 255.255.255.0  broadcast 10.0.2.255
        inet6 fe80::a425:5276:84e3:eb0c  prefixlen 64  scopeid 0x20<link>
        ether 08:00:27:06:03:76  txqueuelen 1000  (Ethernet)
        RX packets 30654  bytes 6052126 (6.0 MB)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 22623  bytes 3558438 (3.5 MB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0

22_pipeslab/pipeslab.txt
vitalipe: damn.. you caught me :P I watched most of them.. but not this one, also you're the first one who mentioned it :)

22_pipeslab/pipeslab.txt
vitalipe: that's what I get on my machine:
```
vitali@Tux:~/tmp$ ifconfig 
lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:420957 errors:0 dropped:0 overruns:0 frame:0
          TX packets:420957 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1 
          RX bytes:142306734 (142.3 MB)  TX bytes:142306734 (142.3 MB)

wlp1s0    Link encap:Ethernet  HWaddr 48:45:20:2d:e9:bb  
          inet addr:10.0.0.11  Bcast:10.0.0.255  Mask:255.255.255.0
          inet6 addr: fe80::8a94:1e71:55d1:45cb/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:413878922 errors:0 dropped:0 overruns:0 frame:0
          TX packets:177838135 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:584336617718 (584.3 GB)  TX bytes:17583095780 (17.5 GB)

vitali@Tux:~/tmp$ cat /proc/version
Linux version 4.4.0-53-generic (buildd@lcy01-28) (gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) ) #74-Ubuntu SMP Fri Dec 2 15:59:10 UTC 2016
```

```ifconfig``` is a horrible command when it comes to parsing it's output.. it's very inconsistent between unix systems, and even between distros...

22_pipeslab/pipeslab.txt
vitalipe: maybe.. we will look into it :)

22_pipeslab/pipeslab.txt
vitalipe: (when it's a directory ```d``` is the first char.. so just ```grep  "^-"```)

vitalipe:
@dovron I'm closing this, if you have any more questions, let me know! I left many comments, but it looks like you got the important things right! just remember:
+ ```wc``` supports flags
+ always be careful with ```cut```
+ a file owner is dictated by the info in the file system, not the current user..

:)

Issue #68, opened 2017-01-02
Recursive Glob - "globstar" 


vitalipe:
There is a cool and almost hidden feature of bash (4 and up) called "globstar" that allows 
you to create recursive globs, like this:
```
**  # will match any file under current dir & it's sub directories recursively

**.c # will match any .c file under current dir & it's sub directories recursively

*/**.c # will match any .c file under any sub directory recursively ,
       #  but in the current directory

```

just remember that you need to enable it first with the following command: 
``` shopt -s globstar```

In the real world you will often use ```find``` for matching files recursively, but still, it's useful to know about it, because newer shells like zsh have it enabled by default, and many build tools use it a lot (this is how I found out about it).

PR #79, opened 2017-01-06

ofirlv:


12_glob_lab/globlab.txt
```diff
@@ -4,26 +4,31 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 
+  Ofir>> for some reason the shell is not case sensitive so also upper case letters are matched, same for the next quesiton. 
+  
   2. All files whose name starts with an uppercase letter
-  Answer:
+  Answer: [A-Z]*
 
   3. All files whose name ends in an "a", "e", "i", "o" or "u"
-  Answer:
+  Answer: *[aeiou].*
```
vitalipe, line 15: a file does not have to contain a dot in their name, for example: ```/etc/passwd``` is a simple text file

12_glob_lab/globlab.txt
```diff
@@ -4,26 +4,31 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
-  Answer:
+  Answer: *[!aeiou].*
 
   5. All files with extension ".c" that are in a folder named "src"
-  Answer:
+  Answer: ls src/*.c
 
   6. All files with extension ".c" under any folder
-  Answer:
+  Answer: ls */*.c
```
vitalipe, line 24: correct! there is also a[ hidden recursive glob](https://github.com/tocodeil/unix1-course-examples/issues/68) in bash :)

12_glob_lab/globlab.txt
```diff
@@ -4,26 +4,31 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 
 2. Write a command that would delete all files ending with a ".png" under any subdirectory.
+$ rm */*.png
 
 3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
+Suppose to print the content of Doc directory. as the match should start with upper case letter. Like I mentioned above, for some reason the match is not case sensitive.
```
vitalipe, line 30: yeah good catch! I try to [explain here](https://github.com/tocodeil/unix1-course-examples/issues/15) let me know if that makes sense to you..

12_glob_lab/globlab.txt
```diff
@@ -4,26 +4,31 @@ https://www.tocode.co.il/bundles/unix1/lessons/globlab
...
 
 4. Search in "man ls" for an option to show only directory data so the command "ls _____ D*" should print only the word Doc
 
+$ ls -l -d [D]*
```
vitalipe, line 34: nice :) but it's possible, and a good practice to combine flags:
```ls -ld``` is the same as ```ls -dl``` and ```ls -l -d``` or ```ls -d -l```

12_glob_lab/globlab.txt
vitalipe: in unix directories are also files.. so the question was a bit vague, there is no glob that I know of that can match files by type.. there is a command called ```file``` that can do that for you, and there are other ways like using ```find``` ...

vitalipe:
@ofirlv -  good job! :) looks like you got the important parts correct. hopefully after reading my comments, the rest is more clear now. If you have any questions feel free to open an issue or mail me :)

PR #80, opened 2017-01-06
ToCode Solution: תרגול find

dovron:


24_find_lab/findlab.txt
```diff
@@ -1,24 +1,65 @@
...
+	
   -. All files and directories not owned by you in /tmp folder
 
+	find /tmp \! -user tocode
+  
   -. All files, folders and links in your home folder whose name ends with a ".c"
-
+	
+	find . -name '*.c'
+	or find /home -name '*.c'
```
vitalipe, line 14: ```/home``` is not your home dir, it's where all the home dirs (except for root) are located by default.

root's home dir is by defaut at ```/root``` this is done to prevent all kinds of issues with permissions and more importantly to prevent denial of service.. it's very common (and a good practice) to mount ```/home``` on it's own partition, that way if a user will fill up that drive, root's home dir will stay unaffected

vitalipe:
@dovron wow! awesome job! you're a true ```find -wizard! -now``` ;)

PR #81, opened 2017-01-06

yanivmix:


30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,18 @@
 Use grep to print only the lines from "input" for which:
 
 1. There is a binary number in the line
-
+grep -E [10] input.txt
```
vitalipe, line 4: it's a good start, but not enough by itself, consider this:
```
vitali@Tux:~/tmp$ echo 943423400112353 | grep [10]
943423400112353
```
we need to match words, there is ```\b```[ that will match words ](http://www.regular-expressions.info/wordboundaries.html)

try:
```grep -E "\b[10]+\b" # the + means match this more then once```

also note that ```-E``` is needed only when using extended regex, it was redundant with only ```[10]```

File: 30_regexp_lab/relab.txt [hunk 2]
```diff
@@ -1,17 +1,18 @@
...
 2. There is an even binary number in the line
-
+grep -E '0$' input.txt
```
vitalipe, line 6: same as above, let's use ```\b```:
```  grep -E "\b[10]+0\b"```
this will work most of the time, except for the number 0, because it will be matched by the ```[10]``` part and ```0``` will not be matched.

we can fix it by using an "or" represented by the pipe symbol:
 ```(A|B) # match either A or B```
like this:
```grep -E "\b([10]+0|0)\b"```

File: 30_regexp_lab/relab.txt [hunk 3]
```diff
@@ -1,17 +1,18 @@
...
 3. Line includes a hexadecimal number
-
+grep -E '0x' input.txt
```
vitalipe, line 8: same as above it's gonna work for this file, but if you want something more generic you can use the same trick that we used for binary :)

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,18 @@
...
+grep -E '^[a-z]' input.txt
 5. Line is not empty
-
+grep -v ^$ input.txt
 6. Line starts with a lowercase letter and ends in a number
-
+grep -E '^[a-z].*[0-9]$'
 7. Line is longer than 50 characters
-
+grep '.\{50\}' input.txt
```
vitalipe, line 16: technically it's 51 ;)

vitalipe:
@yanivmix really nice job! you got all the important parts right! :)

Issue #73, opened 2017-01-03
awk


everybodyls:
in last question you ask to splits this file into multiple file, 
one for each block

awk -F working on files
please explain how to recognize the block ?
what is the answer to this question

ynonp:
Each block starts with a lowercase letter, so the following may be a good indication:

```
/^[a-z]/ { print "Block Start: $0" }
```

The other hint you need is awk can print a line to stdout (as presented) or append it to a file. The following awk script prints the first 5 lines from /etc/shells to a new file:

```
awk 'NR <= 10 { print > newfile.txt }'
```

PR #82, opened 2017-01-07
ToCode Solution: תרגול awk

yanivmix:


32_awk_lab/awklab.txt
```diff
@@ -1,11 +1,12 @@
...
-Answer: 
+Answer: awk "{ print $0, "\n"}' /etc/shells
 
 
 2. Write an awk one-liner that prints the file "shells" with a ">" sign at the beginning of each line
-Answer: 
+Answer: awk '{ print ">"$0}' /etc/shells
 
 3. Write an awk one-liner to swap the first and last words in the output of "help cd"
+awk '{temp = $1, $1 = $NF, $NF = temp}' /bin/cd
```
vitalipe, line 9: 3 things to note here:
* you need to replac ```,``` with ```;```
* you need to pipe the output of the command ```help cd```
* and you need to add the ```print``` action.

something like this should work:
``` help cd | awk '{temp=$1; $1=$NF; $NF=temp}; {print}'```

Note that I added a print action at the end, ```awk``` will look for ```pattern {action}``` pairs, if pattern is omitted the action that flows  will match anything. If  {action} is omitted it will default to {print}, so may times you will see people write it like this:
```help cd | awk '{temp=$1; $1=$NF; $NF=temp};1'```

```1``` is a pattern and the missing action will default to ```{print}```

File: 32_awk_lab/awklab.txt [hunk 2]
```diff
@@ -21,6 +22,6 @@ p2p0: flags=8843 mtu 2304
 
 Write an awk one liner that splits this file into multiple file, one for each block, so after running the script we'll have a file named "en3" with the lines of the first block, and another file named "p2p0" with the lines of the second block.
 
-
+awk '{print $0 " "> "file" NR}' RS='-\\|' /sbin/ifconfig
```
vitalipe, line 25: You first need to match the block - [here](https://github.com/tocodeil/unix1-course-examples/issues/73) Ynon gave some great hints. Here's another one:
``` ifconfig | awk '/^[a-z]/ {file_name=$1 }  {print file_name ":"  $0}'```

vitalipe:
@yanivmix - I'm leaving this open for a while in case you want solve the last question, If you have any questions etc, open an issue :)

PR #85, opened 2017-01-07
ToCode Solution: תרגול הרשאות

LiatKotigaro:


15_perms_lab/permslab.txt
```diff
@@ -1,15 +1,21 @@
...
+  chmod 564 include/player.h
 
   - change permission of all files in Doc folder (without changing the permissions of Doc folder itself) to rw-------
+  chmod 600 Doc/*
 
   - change permission of "lib" folder and all files and subdirectories inside it to rwxrwxrwx
+  chmod -R 777 lib
 
 2. create a new directory named "test" and inside an empty file named "secret". Change permissions of the file "secret" to ---------. Can you delete the file secret? can you delete the folder test? why?
+I was able to delete the file and directory cause I was the owner.
```
vitalipe, line 16: That's a good guess! but it's wrong. You were able to delete "secret" because you have write permissions to it's **parent** directory, try this:
```
vitali@Tux:~/tmp$ touch secret
vitali@Tux:~/tmp$ sudo chmod 000 secret 
vitali@Tux:~/tmp$ sudo chown root.root secret 
vitali@Tux:~/tmp$ ls -lh secret 
---------- 1 root root 0 ינו  8 07:51 secret
vitali@Tux:~/tmp$ rm -f secret 
vitali@Tux:~/tmp$ ls -lh secret 
ls: cannot access 'secret': No such file or directory
```

The (very important) lesson here is that every time you consider the permissions of a file, you should also consider it's parent! 

(It's also the case with files/directories that are read-only inside directories that are writable..)

vitalipe:
@LiatKotigaro good job! :) just make sure you take time to understand the permission parent thing I mentioned in my comment.. it can be very confusing at first, but it's very important! :)

PR #88, opened 2017-01-09

LiatKotigaro:


22_pipeslab/pipeslab.txt
```diff
@@ -1,21 +1,23 @@
 For each of the following, type a single command line that performs:
 
   1. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner.
-  Answer: 
+  Answer: ls -l | cut -c4 | grep x | wc | tr -s ' ' | cut -d" " -f2
```
vitalipe, line 4: you can pass flags to ```wc``` to control it's output. for example with the ```-l``` flag it will only print line count, then we can just:
```ls -l | cut -c4 | grep x | wc -l``` :)

File: 22_pipeslab/pipeslab.txt [hunk 2]
```diff
@@ -1,21 +1,23 @@
...
 
   2. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner, group and others.
-  Answer: 
+  Answer: ls -l | cut -c4,7,10 | grep xxx | wc | tr -s ' ' | cut -d" " -f2
```
vitalipe, line 7: same as above - ```wc -l``` :)

22_pipeslab/pipeslab.txt
```diff
@@ -1,21 +1,23 @@
...
 
   3. Print the number of lines from "ls -l" which represent paths that were modified today (no matter which day we run the command)
-  Answer: 
+  Answer: ls -l --time-style=full-iso| grep $(date +%Y-%m-%d) | wc | tr -s ' ' | cut -d" " -f2
```
vitalipe, line 10: using ```--time-style``` is actually pretty clever :) you can also 
grep on: ```"$(date +%b)\s* $(date +%e)"```

22_pipeslab/pipeslab.txt
```diff
@@ -1,21 +1,23 @@
...
 
   4. Modify (1) and (2) to count files only (no directories)
   Answer:
+  ls -l | cut -c1,4 | grep -x | wc | tr -s ' ' | cut -d" " -f2
+  ls -l | cut -c1,4,7,10 | grep -xxx | wc | tr -s ' ' | cut -d" " -f2
```
vitalipe, line 15: I assume that ```grep -xxx``` & ```grep -x``` are copy-paste errors?
you can pass regex to grep or you can use grep with ```-v``` flag

22_pipeslab/pipeslab.txt
```diff
@@ -1,21 +1,23 @@
...
 
   5. The command /sbin/ifconfig shows information regarding the machine's network interfaces. Use the command to print only the IP address of the machine. Remember not to show the special loopback address 127.0.0.1
-
+/sbin/ifconfig | grep "inet " | grep -v 127.0.0.1 | tr -s ' ' | cut -d" " -f3
 6. Joe wanted to sort a text file in place so he tried to run the command "sort myfile.txt > myfile.txt". What happened to the file? Why?
-
+The content was deleted. I'm not sure what is the reason. Maybe sharing violation (trying writing while reading).
```
vitalipe, line 20: The reason is actually much simpler - the shell will process the redirection first, and by the time ```sort``` is running the file is already empty..

vitalipe:
@LiatKotigaro well done! :) If you have any more questions feel free to open an issue or PM me :)

PR #92, opened 2017-01-14
ToCode Solution: תרגול הרשאות

ofirlv:


15_perms_lab/permslab.txt
```diff
@@ -1,15 +1,19 @@
 1. Using the file structure in the current directory, write the commands that perform the following:
 
   - change permission of file src/main.c to r--r--r-- using chmod letter notation
-
+$ chmod u=r,g=r,o=r src/main.c
   - change permission of file include/player.h to r-xrw--r- using chmod octal (numbers) notation
-
+$ chmod 564 include/player.h
+Ofir>> shouldn't it be r-xrw-r--?
```
vitalipe, line 7: yeah :) three is a typo here, good catch :)

15_perms_lab/permslab.txt
```diff
@@ -1,15 +1,19 @@
...
   - change permission of all files in Doc folder (without changing the permissions of Doc folder itself) to rw-------
-
+$ chmod 600 Doc/*
   - change permission of "lib" folder and all files and subdirectories inside it to rwxrwxrwx
-
+& chmod 777 lib lib/**
```
vitalipe, line 11: you can pass the flag ```-R``` to ```chmod``` which means recursive and it will change all permissions for the directory and  it's content (also deeply nested content):```chmod -R 777 lib```

15_perms_lab/permslab.txt
```diff
@@ -1,15 +1,19 @@
...
 2. create a new directory named "test" and inside an empty file named "secret". Change permissions of the file "secret" to ---------. Can you delete the file secret? can you delete the folder test? why?
+yes, I used (rm test/secret) and got a question if I'd like to delete write protected file. after that the directory got deleted with rm -d
+I'm guessing that it is due to the fact that I'm the owner of the file.
```
vitalipe, line 14: It's a good guess, but no! and that's very important to know why. 
You can delete the file because you have writing permissions to the the parent directory of the file.  

Directories are also files, a very special type of file, that contain a list of file names and their inode. So to delete a file you need to write to the parent directory.  

The practical and the very important lessons here are the following:
 * to know what you can do to a file, you must also consider the parent directory.
* if you give write permissions to a root folder, you effectively give write permissions to the entire tree! (In the final exercise we planted a scenario with this ;) )    

As for the ownership, it only affects the permissions that will be used (owner, group, other)..

vitalipe:
@ofirlv - good job! :) hopefully the comment about the permission thing makes sense, I know that the course is technically over, but feel free to PM me if you have any questions :)

ofirlv:
Thanks.



So, should I stop uploading solution to exercises?

vitalipe:
@ofirlv - if you need more time you can upload stuff here until Friday :)

PR #93, opened 2017-01-14

LiatKotigaro:


30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,25 @@
 Use grep to print only the lines from "input" for which:
 
 1. There is a binary number in the line
+	grep -E -w '[10]+' input.txt | grep -E -v '[10]+[),]'
```
vitalipe, line 4: hmm interesting idea with the 2nd grep! I  but if you only want to match binary numbers without garbage in between, I would simply test for spaces: ``` grep -E  '\s[10]+\s'``` but then:
```
vitali@Tux:~$ echo "111 crap111 c001rap 001" | grep -E  '\s[10]+\s'
``` 

to fix it we can use logical or in the regex, for example: ```(cond1 | cond2)``` will match cond1 OR cond2.. we basically want it be space padded or at the end of the line(&) or at the beginning (^):
```
vitali@Tux:~$ echo "111 crap111 c001rap 001" | grep -E  '(^|\s)[10]+($|\s)'
111 crap111 c001rap 001
```

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,25 @@
...
 
 2. There is an even binary number in the line
-
+	grep -E -w '[10]*0' input.txt | grep -E -v '[10]*0[),]'
+	
 3. Line includes a hexadecimal number
+	grep -E -w -i '0*x*[1-9a-f]+' input.txt | grep -v '^1'
```
vitalipe, line 10: This regex will match ```00000xxxxxxxxxxff``` and it will miss ```0xff0f`` or any hex number with a 0...  (note that 0 is a valid hex digit)

If you want to match hex numbers with hex prefix (0x), I recommend using the same  trick that we used for binary, just adding "0x" at the beginning:``` grep -E '(^|\s)0x[0-9,a-f]+($|\s)' ```

Also note that this pattern will not match all hex numbers in a line, because once it matches a pattern, it "eats" the following white-space.. but we don't really care, we only want to count lines :)

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,25 @@
...
 
 4. Line starts with a lowercase letter
-
+	grep -E '^[a-z]+' input.txt
```
vitalipe, line 13: that's correct! :) but the ```+``` here is a bit redundant, because you only care about the first char

vitalipe:
@LiatKotigaro great job! :)

PR #94, opened 2017-01-14
ToCode Solution: תרגול awk

LiatKotigaro:


32_awk_lab/awklab.txt
```diff
@@ -1,11 +1,12 @@
...
-Answer: 
+Answer: cat /etc/shells | awk ' // { print $0, "\n" } '
 
 
 2. Write an awk one-liner that prints the file "shells" with a ">" sign at the beginning of each line
-Answer: 
+Answer: cat /etc/shells | awk ' // { print "> ", $0 } '
 
 3. Write an awk one-liner to swap the first and last words in the output of "help cd"
+Answer: help cd | awk ' { printf("%s", $NF) } { for(i=2; i<NF; i++) printf(" %s", $i) } { if(NF > 1) printf(" %s", $1) } { printf("\n") } '
```
vitalipe, line 9: Note that```$NF``` will give the value of the last column (technically it will give you the value in column number ```NF```), and ```$1``` will give you the value of the first column. 

So we can just swap them:
```help cd | awk '{temp=$1; $1=$NF; $NF=temp}; {print}'```

File: 32_awk_lab/awklab.txt [hunk 2]
```diff
@@ -20,6 +21,7 @@ p2p0: flags=8843 mtu 2304
     status: inactive
 
 Write an awk one liner that splits this file into multiple file, one for each block, so after running the script we'll have a file named "en3" with the lines of the first block, and another file named "p2p0" with the lines of the second block.
+Answer: /sbin/ifconfig | awk ' BEGIN { block=1 } { if(length($0) > 0) print $0 > block } { if(length($0) == 0) block++ } '
```
vitalipe, line 24: hmm.. almost! you need to make sure the files are named after the the interface, that's actually alot easier than it looks:
```awk '/^[a-z]/ {file_name=$1 } {print $0 >> file_name}'```


```/^[a-z]/ {file_name=$1 }``` - match any line that starts with lower case letter and set ```file_name``` to the first word for that line

```{print $0 >> file_name}``` - match **everything**, and print it to a file called ```file_name```

if it's not clear, try this:
```ifconfig | awk '/^[a-z]/ {file_name=$1 } {print file_name ":" $0}'```

vitalipe:
@LiatKotigaro  - good job! :) hopefully my comments helped to clear the confusion a bit. awk is a hard tool to master! it's almost as user hostile as vim :P

PR #99, opened 2017-05-11

Yulia88:


07_cmd_lab/cmdlab.txt
```diff
@@ -1,15 +1,43 @@
...
+bash
 
-2. View the file /etc/passwd on your machine with the command "cat /etc/passwd". Then read about the file in "man 5 passwd". What information is held in the file? What are the fields in each line?
+2. View the file /etc/passwd on your machine with the command "cat /etc/passwd".
+ Then read about the file in "man 5 passwd". What information is held in the file? What are the fields in each line?
 
-3. In Unix, the /etc folder holds configuration files. Display all files in "/etc" folder ordered by file size (Hint: search in "man ls")
+it's a password file. holds per each user one line and 7 fields :
+login name 
+optional encrypted password
```
ynonp, line 28: Optional only in theory :) In practice you won't find anyone's encrypted password there for the past 30 years or so. Modern practice keeps all users encrypted passwords in the file `/etc/shadow`

07_cmd_lab/cmdlab.txt
```diff
@@ -1,15 +1,43 @@
...
+user name or comment field
+user home directory
+optional user interpreter 
+
+
+3. In Unix, the /etc folder holds configuration files. Display all files in "/etc" folder ordered by file size (Hint: search in "man ls")
+
+
+ls -l --block-size /etc/
+
```
ynonp, line 39: Did it work? I tried and got the error:

```
 invalid --block-size argument '/etc'
```

block size controls how ls prints the size column, for example it lets you print the size in bytes, kilo-bytes, MB etc. 

your task is to find another option that sorts according to the file size, but also as a bonus you can try to get block-size to work

07_cmd_lab/cmdlab.txt
```diff
@@ -1,15 +1,43 @@
...
 
 4. Print current date/time formatted as: "Sun 21/03/2010, 14:30". Hint: read "man date" and search for the correct format string.
+
+date +'%a %m/%d/%Y, %R'
```
ynonp, line 43: 👍

ynonp (commented):
Hi Yulia,
Great work! Now you only need to find how to get `ls` to sort files by size and move on to the next exercise

07_cmd_lab/cmdlab.txt
Yulia88: I mistakenly wrote the wrong command
Block size: 
ls -l --block-size=M /etc/

The command that was in the question is:
ls -l -S /etc/

ynonp:
Fantastic. Closing.

PR #104, opened 2017-05-20

Yulia88:


12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,59 @@
...
+ 
+ls -d [[:lower:]]*/*
+ 2. All files whose name starts with an uppercase letter
   Answer:
 
-  3. All files whose name ends in an "a", "e", "i", "o" or "u"
+ 
+ls -d [[:upper:]]*/* - for some reason shows the following:
+Doc/README.txt Doc/test.c
+Why test.c as well?
```
ynonp, line 20: The problem is with the forward slash `/`:

Your glob pattern matches all the files in all directories which (the directory) start with an uppercase letter.

Better to skip the second part of the glob and leave it at:

```
ls -d [[:upper:]]*
```

Or, if you also want to see the contents of each directory drop the d:

```
ls [[:upper:]]*
```

Alternatively, showing all the FILES that start with an uppercase in all directories would look like this:

```
ls -d */[[:upper:]]*
```

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,59 @@
...
+
+ 3. All files whose name ends in an "a", "e", "i", "o" or "u"
   Answer:
 
-  4. All files whose name DOES NOT end in an "a", "e", "i", "o" or "u"
+ 
+echo *[a,e,i,o,u]
```
ynonp, line 26: Better without the comma: `*[aeiou]`. A comma is a character like any other and so a file named `hi,` will match your pattern

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,59 @@
...
+ 
+echo *[!aeiou]
+
+ 5. All files with extension ".c" that are in a folder named "src"
   Answer:
 
-  6. All files with extension ".c" under any folder
+
+
+ls *.c src/
```
ynonp, line 39: That doesn't work. It shows .c files from current directory AND all files in src directory. You can try and create a `main.c` file in the current directory and a `demo.h` file in src to see the problem

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,59 @@
...
+  
+6. All files with extension ".c" under any folder
   Answer:
 
-2. Write a command that would delete all files ending with a ".png" under any subdirectory.
+
+find -name '*.c'
```
ynonp, line 45: That works. But you can also solve it using glob patterns. Do give it a second thought

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,59 @@
...
+
+     2. Write a command that would delete all files ending with a ".png" under any subdirectory
+find -name '*.png* -delete
```
ynonp, line 48: Again- works, but I'd rather see a solution using glob patterns

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,59 @@
...
+
+     
 
-3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
+3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
+Printed same thing as in answer to question 2, I don't understand why only these files are printed 
+
```
ynonp, line 54: I gave some hints there. Let me know if it helped.

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,59 @@
...
+     
 
 4. Search in "man ls" for an option to show only directory data so the command "ls _____ D*" should print only the word Doc
 
+ls -d D*
```
ynonp, line 59: 👍

ynonp (changes requested):
Hi Yulia,

Great work all in all. I loved your searching around for solutions even outside lesson materials (eg. using find).

I suggested some improvements and you still have one answer that doesn't work and needs fixing.

12_glob_lab/globlab.txt
Yulia88: OK, got it, thanks!

12_glob_lab/globlab.txt
Yulia88: I have the same output for both, I don't understand the difference, can you please explain?

12_glob_lab/globlab.txt
Yulia88: the only way I succeed is by actually being inside the src dir
capture
And I have a feeling i can use here ls, but I tried everything we learned and it doesn't work..

12_glob_lab/globlab.txt
Yulia88: Was printed everything within the upper case folders which is Doc/ 
But how do I show only files with upper case in only Doc

12_glob_lab/globlab.txt
Yulia88: Finally!!
files

As well, question
If I created file.png in current folder "wildcards" and also *.png files in img/ and Doc/ 
Why the ls command can't show me all three png files, but only two
capture2

ynonp: 👍 
(though you can do just fine without the square brackets - `ls -d */*.c`)

12_glob_lab/globlab.txt
Yulia88: rm -d */*[.png]

Yulia88 (commented):
I have added fixes

12_glob_lab/globlab.txt
ynonp: Try this:

```
touch xyz, xa
echo *[aeiou]
echo *[a,e,i,o,u]
```

On my machine this is the output:

```
# for echo *[aeiou]:
file xa

# for echo *[a,e,i,o,u]
file xa xyz,
```

Different outputs when a file ends in a comma

Yulia88: Got it, so basically the comma as you explained is a character and it will find such files if they exist

12_glob_lab/globlab.txt
ynonp: I assume you mean - `rm */*.png`

1. No `-d` for rm
2. Square brackets around a word is never a good idea

12_glob_lab/globlab.txt
ynonp: Not sure I understand - what command are you running and what output do you expect to see?

12_glob_lab/globlab.txt
ynonp: Re. your question - please also attach the output of `tree` so we can see which files you actually have in that folder

12_glob_lab/globlab.txt
Yulia88: Yes, i can't print "*" in here for some reason 
1. ok
2. ok

12_glob_lab/globlab.txt
Yulia88: Still have an issue with installing the tree
I have responded in the "tree" question

12_glob_lab/globlab.txt
Yulia88: I meant that I understood the answer ( I think) 
the command ls -l [A-Z]* prints out everything within the upper case folder which is Doc

What I asked was how do I print out upper case files only in Doc folder only

12_glob_lab/globlab.txt
Yulia88: Attached snapshot 
tree

Yulia88: Here is the snapshot
In overall I have 4 png files 
ls command shows only 3 png although there are 4 
tree

ynonp: Command should be `ls */*.png` (we'll talk about the reasons tomorrow if it's still not clear)

PR #106, opened 2017-05-22

Gadpal:


12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,17 @@
...
-
-4. Search in "man ls" for an option to show only directory data so the command "ls _____ D*" should print only the word Doc
-
+1. Mkdir  wildcards
+2. Cd wildcards � mkdir� (all folder listed
+3. Touch command for files.
+4. 
+a. echo [a-z]* (using this command export LC_ALL=POSIX)
+b. echo [A-Z]*
+c. echo * [a,e,i,o,u]
```
ynonp, line 7: The space is an error, and commas are not required. Correct answer is: `echo *[aeiou]`

(comma is a normal character, and not required for separation. The space meant you'll get "All Files" AND "All files that are the single letter a,e,i,o or u", whereas what you wanted is the all files that end in a vower)

Spaces in shell are very important as they function as word separators.

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,17 @@
...
+d. echo * [!a,e,i,o,u]
```
ynonp, line 8: Same as in previous question - space is an error and correct answer is `*[!aeiou]`

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,17 @@
...
+e. cd src => echo *.c
+f. could not find other way (same as the above: cd dir => echo *.c)
```
ynonp, line 10: The trick here is you can use a `/` (directory separator) inside a glob pattern. So the following works: `echo src/*.c`

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,17 @@
...
+g. could not find a way with one command line � not sure I got this task right.
```
ynonp, line 11: Yes once you realize you can use a `/` inside glob patterns, it's easier. For example all c files in all folders would be `echo */*.c`.

All png files in all folders would be `*/*.png`

All files that start with a `d` in folders that start with an uppercase letter would be: `echo [A-Z]*/d*`

etc.

ynonp (approved):

PR #112, opened 2017-05-27
ToCode Solution: תרגול הרשאות

roigizmo:


15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,40 @@
...
-  - change permission of file include/player.h to r-xrw--r- using chmod octal (numbers) notation
-
-  - change permission of all files in Doc folder (without changing the permissions of Doc folder itself) to rw-------
-
-  - change permission of "lib" folder and all files and subdirectories inside it to rwxrwxrwx
-
-2. create a new directory named "test" and inside an empty file named "secret". Change permissions of the file "secret" to ---------. Can you delete the file secret? can you delete the folder test? why?
+ 
+
+chmod a-wx src
```
ynonp, line 7: Works. But I think `chmod a=r src/main.c` is more readable. (Also I think you forgot the ending main.c)

15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,40 @@
...
+
+ - change permission of file include/player.h to r-xrw--r- using chmod octal (numbers) notation
+
+  
+
+562 include
```
ynonp, line 13: Yes. Or in its full version `chmod 562 include/player.h`

15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,40 @@
...
+
+- change permission of all files in Doc folder (without changing the permissions of Doc folder itself) to rw-------
+
+  -
+
+rchmodf a-x go-rwx Doc
```
ynonp, line 19: That would change permissions on Doc folder itself, which is not what the question asked. We need to change the permissions of the files inside Doc, but not of the folder.

The rationale is a directory MUST have execute permission. Without it you can't do anything in that directory.

So correct answer here is `chmod 600 Doc/*`

15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,40 @@
...
+
+
+ change permission of "lib" folder and all files and subdirectories inside it to rwxrwxrwx
+
+2
+
+chmod -R go+w lib
```
ynonp, line 26: 👍

15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,40 @@
...
+
+. create a new directory named "test" and inside an empty file named "secret". Change permissions of the file "secret" to ---------. Can you delete the file secret? can you delete the folder test? why?
+
+
+i can't delete the file and the folder. i do not have any premission to read, write or execute.
```
ynonp, line 31: What commands did you use? What was the output?

15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,40 @@
...
+i can't delete the file and the folder. i do not have any premission to read, write or execute. 
+
+
 
 3. Read the output of "help umask". Explain in your own words what umask does, and how would you use it to set new files permissions to ------r--
 
 
+
+
+unmask o+r
```
ynonp, line 40: That wouldn't work. Try to run the following, and after each create some new files and check the permissions of the created files:

```
umask 000
umask 111
umask 444
umask 600
umask 666
umask 777
```

Then try to find how umask works accordingly

ynonp (changes requested):
Please correct last 2 questions

PR #122, opened 2017-06-02

Yulia88:


22_pipeslab/pipeslab.txt
```diff
@@ -1,34 +1,78 @@
...
 
-  1. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner.
+ 
+ 1. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner.
   Answer: 
 
-  2. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner, group and others.
+ 
+
+ls -l | cut -d- -f1 |grep d |wc -l
```
ynonp, line 9: This prints only directories. You can do better with `cut -c` which prints specific characters, so this one is more precise: `ls -l | cut -c4 | grep -c x

22_pipeslab/pipeslab.txt
```diff
@@ -1,34 +1,78 @@
...
+ls -l | grep x" " |wc -l
+
+ 
+3. Print the number of lines from "ls -l" which represent paths that were modified today (no matter which day we run the command)
   Answer: 
 
-  4. Modify (1) and (2) to count files only (no directories)
+ 
+
+ls -l |grep Jun | grep -n 2 | wc -l
```
ynonp, line 29: What is a file name contains Jun 2, for example try `touch "June 2"` ?

22_pipeslab/pipeslab.txt
```diff
@@ -1,34 +1,78 @@
...
-What was printed? Why? Fix the command line to get the output:
+What was printed? 
+Why? the -f isn't correct, it shows an empty space 
+
+Fix the command line to get the output:
 2 Dec
 2 Nov
+
+
+who | cut -d" " -f14 |sort|uniq -c
```
ynonp, line 76: This works, but we can get better results with `cut -c`

ynonp (commented):
You're on the right track but I feel some direction is needed. I'm thinking about another Q&A session on this subject. I'll email

PR #133, opened 2017-06-04

moshikob99:


12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,30 @@
...
-
-4. Search in "man ls" for an option to show only directory data so the command "ls _____ D*" should print only the word Doc
-
+Glob Lab.
+https://www.tocode.co.il/bundles/unix1/lessons/globlab
+
+1. Type in a glob pattern to match:
+
+  1. All files whose name starts with a lowercase letter
+  Answer: ls [a-z]* or ls [[:lower:]]
```
ynonp, line 7: 👍

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,30 @@
...
+
+  2. All files whose name starts with an uppercase letter
+  Answer: ls [A-Z]*????
```
ynonp, line 10: Correct. Why the question marks?

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,30 @@
...
+       or ls [[:upper:]]
+
+  3. All files whose name ends in an "a", "e", "i", "o" or "u"
```
ynonp, line 13: 👍

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,30 @@
...
+  Answer:ls *[aeiou]
+
+  4. All files whose name DOES NOT end in an "a", "e", "i", "o" or "u"
+  Answer:ls *[!aeiou]
```
ynonp, line 17: 👍

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,30 @@
...
+
+  5. All files with extension ".c" that are in a folder named "src"
+  Answer:ls *.c /src
```
ynonp, line 20: This doesn't work. You called ls with the list of `.c` files in current directory AND the word `/src`. What it'll do is show all `.c` files in current directory and contents of `/src` folder if exists (but probably an error message since it doesn't usually exists)

What you want to do here is combine the two to a single glob pattern stating all files in src folder that ends in .c. The command is therefore: `ls src/*.c`

Note also that `/src` and `src` refer to 2 different directories: The first is at the top of the filesystem tree, and the second is relative to your current working directory. So if you're in `$HOME/lab` the second is actually `/home/moshiko/lab/src`, while the first is always just `/src` no matter from where you use it

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,30 @@
...
+
+  6. All files with extension ".c" under any folder
+  Answer: >>>> ls *.c ??????
```
ynonp, line 23: This is a continuation of the previous question-

Now that you know that you can use `/` in a glob pattern to separate directories, and `src/*.c` is all c files in src you can replace `src` with star to get:

`ls */*.c` which means all .c files in all folders

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,30 @@
...
+
+2. Write a command that would delete all files ending with a ".png" under any subdirectory.
+  rm *.png ?????
```
ynonp, line 26: And the final question in that series changes .c to .png to get `rm */*.png`

12_glob_lab/globlab.txt
```diff
@@ -1,29 +1,30 @@
...
+
+3. The command "ls -l" can take the name of a file or directory and show details on them. From within "wildcards" folder run "ls -l [A-Z]*". What was printed? Why?
+
+4. Search in "man ls" for an option to show only directory data so the command "ls _____ D*" should print only the word Doc
```
ynonp, line 30: When I ran the commands I got the following result:

```
$ ls [A-Z]*
README.txt test.c
```

I think it's surprising because one would expect to find only files starting with an uppercase, and then there's test.c in the list.

The reason is the glob pattern actually expanded to the single entry `Doc`, so the command executed was in fact:

```
$ ls Doc
```

This is because writing `[A-Z]*` in wildcards folder results in only one match, the name `Doc`

Now `ls` itself, when called with a directory name will show all files in that directory. That's why we got everything under Doc. You can verify it by creating more files in Doc and you can see they will also be printed.

One solution if you do need to see only the glob result is to add `-d` to ls. That flag causes ls not to print directory contents and so the result is:

```
$ ls -d [A-Z]*
Doc
```

ynonp (approved):
Hi Moshiko,
Good work. I added clarifications and explanations where needed. Please review them and let me know if you have any further questions

PR #143, opened 2017-06-07

BENIDUDE:


22_pipeslab/pipeslab.txt
```diff
@@ -2,20 +2,44 @@ For each of the following, type a single command line that performs:
 
   1. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner.
   Answer: 
+	ls -l|grep $USER| cut -c 4|grep x|wc -w
+	[Q: Based on the commands we know by now is there a shorter way to do 	this? With less pipes?]
```
ynonp, line 7: You don't need the first `grep`, as the question is not restricted to YOUR files. And one nice shortcut is to use `grep -c` (which just prints a count). Therefore you can use: `ls -l | cut -c 4 | grep -c x`

22_pipeslab/pipeslab.txt
```diff
@@ -2,20 +2,44 @@ For each of the following, type a single command line that performs:
 
   1. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner.
   Answer: 
+	ls -l|grep $USER| cut -c 4|grep x|wc -w
+	[Q: Based on the commands we know by now is there a shorter way to do 	this? With less pipes?] 
 
   2. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner, group and others.
   Answer: 
+	ls -l|grep $USER| cut -c 4,7,10|grep x|wc -w
```
ynonp, line 10: Searching for a single `x` is not sufficient, for a line may have x on column 4 but `-` in 7 and 10. Better to: `ls -l | cut -c 4,7,10 | grep -c xxx`

22_pipeslab/pipeslab.txt
```diff
@@ -2,20 +2,44 @@ For each of the following, type a single command line that performs:
...
 
   3. Print the number of lines from "ls -l" which represent paths that were modified today (no matter which day we run the command)
   Answer: 
+	
+	ls -l | cut -f6-8 -d" "| grep $(date +%b" "%d)| wc -l
```
ynonp, line 15: 1. cut is not good at handling duplicate delimiters, so if you want to use space as a delimiter you'll need to make sure no duplicate spaces appear on the line. ls -l actually prints a lot of duplicate spaces, so we can use tr to reduce them, i.e.: `ls -l | tr -s ' '`

2. date allows you to control its padding strategy. `date +"%b %-d"` won't pad the day with zeroes

3. You need to use quotes around the $(...)

Combine all and get: `ls -l | tr -s ' ' | cut -d' ' -f6-7|grep "$(date '+%b %d')" | wc -l`

22_pipeslab/pipeslab.txt
```diff
@@ -2,20 +2,44 @@ For each of the following, type a single command line that performs:
...
+	Below an answer where I used a set date for today:
+	ls -l | cut -f6-8 -d" "| grep "Jun  7"| wc -l (working)
+	]
 
   4. Modify (1) and (2) to count files only (no directories)
   Answer:
 
+	(1) ls -l|grep $USER|cut -c 1-4|grep x |cut -c1 | grep - | wc -l
+	(2) ls -l|grep $USER|cut -c 1,4,7,10|grep x| cut -c1 | grep -|wc -w
+
```
ynonp, line 31: A nice tip here is that you can actually use `grep -x`, but you need to tell grep that the `-x` is what you're looking for and not a search option. You do that by passing `--`, which is a special option marking the end of dash options and beginning of actual parameters.

So the following shortcut works: `ls -l | cut -c 1,4 | grep -c -- -x`

22_pipeslab/pipeslab.txt
```diff
@@ -2,20 +2,44 @@ For each of the following, type a single command line that performs:
...
+
   5. The command /sbin/ifconfig shows information regarding the machine's network interfaces. Use the command to print only the IP address of the machine. Remember not to show the special loopback address 127.0.0.1
 
+	/sbin/ifconfig | tr -s " " | cut -f3 -d" " | grep addr | head -1
```
ynonp, line 35: This will only print the first line, while in real life there can be more, and order is not always known. Luckily we can use `grep -v` to say "print everything without the word" getting: `/sbin/ifconfig | ... | grep -v 127.0.0.1`

22_pipeslab/pipeslab.txt
```diff
@@ -31,4 +55,15 @@ What was printed? Why? Fix the command line to get the output:
 2 Dec
 2 Nov
 
+who | tr -s " " | cut -d ' ' -f 3| cut -f2 -d" "| cut -f2 -d-| uniq -c
+
+[Note: Below is the ouput I am getting on my machine,
+We can use the tr command to replace for example 05 of May but maybe there is a better why to do it?]  
+
```
ynonp, line 62: This is cause by locale definitions which indicate to `who` you prefer to get its month name as a number. You can fix it with: `LC_ALL=C who | tr -s ' ' | cut -d ' ' -f 3 | sort | uniq -c`

ynonp (commented):
Great work! Do review my comments and suggestions

22_pipeslab/pipeslab.txt
BENIDUDE: 1. In this case I think we had no extra space in the output, That's the reason I didn't use tr
2. Thanks for this tip

PR #145, opened 2017-06-07
ToCode Solution: תרגול הרשאות

Nadav1234:


15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,33 @@
 
   - change permission of file src/main.c to r--r--r-- using chmod letter notation
 
+chmod u=r,g=r,a=r src/main.c
+
   - change permission of file include/player.h to r-xrw--r- using chmod octal (numbers) notation
 
+mews2659 ~/wildcards >> chmod 474 include/player.h
```
ynonp, line 9: 474 is r--rwxr--
The answer here is: `chmod 564 include/player.h`

15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,33 @@
...
 
+chmod u=r,g=r,a=r src/main.c
+
   - change permission of file include/player.h to r-xrw--r- using chmod octal (numbers) notation
 
+mews2659 ~/wildcards >> chmod 474 include/player.h 
+
   - change permission of all files in Doc folder (without changing the permissions of Doc folder itself) to rw-------
 
+mews2659 ~/wildcards >> chmod 600 Doc/*.*
```
ynonp, line 13: Prefer `*` over `*.*` as not all files have a period in them. Better to use `chmod 600 Doc/*`

15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,33 @@
...
+
   - change permission of "lib" folder and all files and subdirectories inside it to rwxrwxrwx
 
+chmod -R 777 lib/
```
ynonp, line 17: 👍

15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,33 @@
...
+
 2. create a new directory named "test" and inside an empty file named "secret". Change permissions of the file "secret" to ---------. Can you delete the file secret? can you delete the folder test? why?
 
+mkdir test
+touch test/secret
+chmod 000 test/secret
+mews2659 ~/wildcards >> rm -r test/
+rm: remove write-protected regular empty file ‘test/secret’? y
+
+It worked because there were no permissions for the file. Perhaps because I am using tcsh?
```
ynonp, line 27: You did good. Actually a more interesting question would be to change permission of the folder `test` to `000`. Then you'd see that no matter what permissions you give to `secret` you won't be able to delete it.

15_perms_lab/permslab.txt
```diff
@@ -2,14 +2,33 @@
...
+
 3. Read the output of "help umask". Explain in your own words what umask does, and how would you use it to set new files permissions to ------r--
 
+umask displays the default permissions newly created files receive.
```
ynonp, line 31: It's the opposite- umask is subtracted from a `rw-rw-rw` full permission and the result is the default permission for new files. So a umask of 002 is calculated as:

```
rw-rw-rw-
-
-------w-
=
rw-rw-r--
```
Which means a umask of `002` results in default permission to files `rw-rw-r--`. Our goal was to reach a default permission of `------r--` so we need to have umask deleting the user and group permissions.

All in all correct answer is: `umask 662`

ynonp (approved):
Please review my comments and suggestions before moving on

PR #146, opened 2017-06-09

Gadpal:


30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,7 @@
...
-4. Line starts with a lowercase letter
-
-5. Line is not empty
-
-6. Line starts with a lowercase letter and ends in a number
-
-7. Line is longer than 50 characters
-
-8. Line is of an even length
+1. grep -E '[0-9]{0,1}' input | head -11
```
ynonp, line 1: This one prints ALL the lines, even the ones without binary numbers. What it means is "print every line that has at least zero or one digit in it", but every line has at least 0 digits...

The following is the correct answer: `grep [01] input`.

Now this prints all lines that contain a 0 or 1 in them, which is nice but will also print a line that has the number 70 (which is not really a binary number).

We can improve with the `-w` option to grep, which means to search for full words. The command `egrep -w [01]+ input` will therefore print only sequences of zeroes and ones that form a full word

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,7 @@
...
+2.
```
ynonp, line 2: `egrep -w '[01]*0'` - see comment on previous answer

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,7 @@
...
-5. Line is not empty
-
-6. Line starts with a lowercase letter and ends in a number
-
-7. Line is longer than 50 characters
-
-8. Line is of an even length
+1. grep -E '[0-9]{0,1}' input | head -11
+2. 
+3. grep hexadecimal input
```
ynonp, line 3: A hexadecimal number is a number that consists of the digits 0-9 and the letters a-f, so to find such sequences we'll use `grep -w [0-9a-f]+`

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,7 @@
...
+4. grep ^[a-z] input
```
ynonp, line 4: With grep you almost always want to 'protect' your regular expression with quotes. Without them you risk bash will treat the word as a glob pattern.

Better to write: `grep '^[a-z]' input`

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,7 @@
...
+5. grep -v -e '^$' input
```
ynonp, line 5: This works, but I like the following shorter version better: `grep . input`, which works because . means any character, and so the expression means "lines that contain at least one any character"

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,7 @@
...
+6. grep '[^a-z] [$0-9]' input
```
ynonp, line 6: 1. In Regular expressions a space does not mean AND but just a space
2. Inside [...] the signs ^ and $ lose their meaning

so your expression translated to words is: "all lines that contain one character that is not a lowercase letter, then a space, and then a single character that is either a $ or a digit."

A correct version looks like this: `^[a-z].*[0-9]$`. Here's why:

1. `^` is the beginning of a line
2. `[a-z]` next comes a lowercase letter
3. `.*` next comes a sequence of characters. `.` means any character and the following `*` means that thing can appear any number of times.
4. `[0-9]` then a single digit
5. `$` and finally the end of the line. Nothing should ever appear after a `$` because no characters can appear after the end of the line

30_regexp_lab/relab.txt
```diff
@@ -1,17 +1,7 @@
...
+7. grep '.\{50\}' input
```
ynonp, line 7: Very nice! Or nicer with egrep `egrep '.{50}' input`

ynonp (approved):
Please review my comments and suggestions before moving on. Let me know if anything is not clear or require further explanations

PR #148, opened 2017-06-09
ToCode Solution: תרגיל Links

moshikob99:


18_links_lab/linkslab.txt
```diff
@@ -1,20 +1,44 @@
...
+
+
+
+2. Move "file3" to the new "~/tmp" directory you just created. Can you still use it to see the original "file1" ? Why not? How would you create a symbolic link that would survive the transfer to ~/tmp ?
+   no we cant use it 
+   ln -s $(realpath file1) file3
+
+
+
+3. How can you tell that "file2" and "file4" refer to the same file?
```
ynonp, line 43: You can use `ls -li` to see the inode number of both files: 2 files are hard links to each other if they have the same inode number

ynonp (approved):
Great Work!

PR #149, opened 2017-06-12
ToCode Solution: תרגיל Links

Nadav1234:


18_links_lab/linkslab.txt
```diff
@@ -9,12 +9,33 @@ Preparation: Perform the following  (no need to write the commands you used):
...
+
 2. Move "file3" to the new "~/tmp" directory you just created. Can you still use it to see the original "file1" ? Why not? How would you create a symbolic link that would survive the transfer to ~/tmp ?
 
+mv file3 /homes/nadavg/tmp/
+cat: file3: No such file or directory
+
+I do not know how to create a symbolic link that would survive the transfer to ~/tmp
+Did you mean using $ln without the -s?
+Is it still symbolic?
+
```
ynonp, line 32: Removing the -s will create a hard link, so that's not the answer.

The key here is that symbolic link is a file that contains the PATH to another file/directory. that path can be absolute or relative.

A symbolic link to a relative path looks like this:
file3 -> file1
And is created with `ln -s file1 file3`
It means file3 will link to a file named `file1` that is in the same folder as file3. Move file3 and you'll need to also move file1 to get them working.

An absolute link looks like this: `file -> /etc/passwd`, i.e. it starts with a `/`. This means no matter where the link is located, it will always point to the same destination.

You create absolute symbolic links by specifying the full path to their target, such as: `ln -s /etc/passwd file`, or in your case `ln -s /home/nadavg/lab/links/file1 file3`

Since the first argument to ln is an absolute path, the result link will be absolute.

BTW a nice shortcut uses the command realpath to automatically generate the full path to a file, so this line works just the same:
`ln -s $(realpath file1) file3`

18_links_lab/linkslab.txt
```diff
@@ -9,12 +9,33 @@ Preparation: Perform the following  (no need to write the commands you used):
...
+
 3. How can you tell that "file2" and "file4" refer to the same file?
+
+~/links$ ls -li
+2153614395 -rw-rw--w- 2 nadavg develop6   5 Jun 12 17:00 file2
+2153614395 -rw-rw--w- 2 nadavg develop6   5 Jun 12 17:00 file4
+
+They have the same serial number.
+
```
ynonp, line 41: 👍

ynonp (approved):
Good work. Please see my explanation on relative and absolute symbolic links here in the comments

PR #155, opened 2017-06-15

roigizmo:


37_environment_lab/envlab.txt
```diff
@@ -3,7 +3,18 @@
...
   - cp => cp -i
 
-2. How can you "install" all defined aliases into the current shell? How can you add them permenantly to all shells?
+
+
+alias ll= "ls -l"
+alias rm= "rm -i"
+alias cp ="cp -i"
+
+2. How can you "install" all defined aliases into the current shell? How can you add them permenantly to all shells?
```
ynonp, line 12: After running the above `alias` commands, leaving the shell and re-entering (for example if you close the window and open a new one) all aliases defined will disappear.

In order to define permanent aliases you need to add the above alias lines to a startup file (also called rc file) named `$HOME/.bashrc`, specifically in your case it's `/home/roi/.bashrc`

ynonp (commented):
Missing answers to the last question.

PR #157, opened 2017-06-17

moshikob99:


22_pipeslab/pipeslab.txt
```diff
@@ -1,34 +1,62 @@
...
-2 Nov
-
-
+For each of the following, type a single command line that performs:
+
+
+
+  1. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner.
+
+  Answer: ls -l | cut -c 4 | grep -c x
```
ynonp, line 7: 👍

22_pipeslab/pipeslab.txt
```diff
@@ -1,34 +1,62 @@
...
+
+
+  1. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner.
+
+  Answer: ls -l | cut -c 4 | grep -c x 
+
+
+  2. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner, group and others.
+
+  Answer:ls -l | cut -c 4,7,10 | grep -c x
```
ynonp, line 12: 👍

22_pipeslab/pipeslab.txt
```diff
@@ -1,34 +1,62 @@
...
+
+  2. Print the number of lines from "ls -l" which represent a path that has execute permission to their owner, group and others.
+
+  Answer:ls -l | cut -c 4,7,10 | grep -c x 
+
+
+
+  3. Print the number of lines from "ls -l" which represent paths that were modified today (no matter which day we run the command)
+
+  Answer: ls -l  cut -f 6,7-d' ' |grep +(date +%b� �%d) wc -l
```
ynonp, line 18: Missing tr - as some lines have multiple spaces the cut doesn't always work in your suggestion. You need to add `tr -s ' '` which removes duplicate spaces.

22_pipeslab/pipeslab.txt
```diff
@@ -1,34 +1,62 @@
...
+
+
+
+  4. Modify (1) and (2) to count files only (no directories)
+
+  Answer:ls -l | cut -f1 | grep -c x
```
ynonp, line 24: To compare only files we need to first filter the lines starting with a `-` (that indicates a file) and then from THAT new list filter the lines that have `x` on their 4th column:

`ls -l | cut -f1 | grep - | cut -c 4 | grep -c x`

22_pipeslab/pipeslab.txt
```diff
@@ -1,34 +1,62 @@
...
+
+
+
+  5. The command /sbin/ifconfig shows information regarding the machine's network interfaces. Use the command to print only the IP address of the machine. Remember not to show the special loopback address 127.0.0.1
+
+    /sbin/ifconfig | cut ...
+
+6. Joe wanted to sort a text file in place so he tried to run the command "sort myfile.txt > myfile.txt". What happened to the file? Why?
+
+   the file myfile.txt was deleted ( run over..)
```
ynonp, line 34: Yes. And it's a bit surprising because one would hope the output of sort will be written into the file, and the result will have that file sorted.

Unfortunately the order of execution in a shell pipeline begins from the end, and so what actually happens:

1. Bash opens the file myfile.txt for writing, truncating its contents
2. Bash runs sort myfile.txt which tries to sort the lines in the file, but the file is already empty so nothing to sort
3. The output of sort is written back to the file, leaving it empty

22_pipeslab/pipeslab.txt
```diff
@@ -1,34 +1,62 @@
...
+
+. 
+
+I want to show how many connections began in each month so I tried to run the command:
+
+$ who | cut -d ' ' -f 3 | sort | uniq -c
+
+ we receved the pts number
+
+What was printed? Why? Fix the command line to get the output:
```
ynonp, line 58: The problem was multiple spaces in the output of `who` made `cut` print wrong responses. We can work around it by adding tr, which would remove the duplicate spaces:

`who | tr -s ' ' | cut -d ' ' -f 3 | sort | uniq -c`

ynonp (approved):
Please see my comments

PR #160, opened 2017-06-22

Jozefff:


30_regexp_lab/relab.txt
```diff
@@ -2,16 +2,33 @@ Use grep to print only the lines from "input" for which:
 
 1. There is a binary number in the line
 
+$ grep -E -w [0,1]+ input
```
ynonp, line 5: 1. Better to use `egrep`
2. No need to put a comma inside square brackets. Simply use `[01]+`

30_regexp_lab/relab.txt
```diff
@@ -2,16 +2,33 @@ Use grep to print only the lines from "input" for which:
...
+
 2. There is an even binary number in the line
 
+grep -E -w [0,1]+[0] input
```
ynonp, line 9: Use `*` here instead of `+` to allow the number 0

30_regexp_lab/relab.txt
```diff
@@ -2,16 +2,33 @@ Use grep to print only the lines from "input" for which:
...
 
+grep -E -w [0-9a-f]+ input
+
 4. Line starts with a lowercase letter
 
+grep -E '^[a-z]' input
+
 5. Line is not empty
 
+grep -E ^. input
```
ynonp, line 21: Or the more compact `egrep . input`

ynonp (commented):
Very good!

PR #163, opened 2017-06-27

moshikob99:


37_environment_lab/envlab.txt
```diff
@@ -1,15 +1,37 @@
...
+
+  - ll => ls -l
+
+  - rm => rm -i
+
+  - cp => cp -i
+    
+    alisa ll="ls -l"
+    alias rm="rm -i"
+    alais cp="cp -i"
```
ynonp, line 11: 👍

37_environment_lab/envlab.txt
```diff
@@ -1,15 +1,37 @@
...
+    echo �a ll 'ls -l'�>> file name ?
+
+2. How can you "install" all defined aliases into the current shell? How can you add them permenantly to all shells?
+
+  we need to check the filr bachrc
+  we need to add the aliases to bach_aliases
```
ynonp, line 17: 👍

37_environment_lab/envlab.txt
```diff
@@ -1,15 +1,37 @@
...
+3. Add the following aliases into the aliases file you defined previously:
+
+  - count_files => print the number of files in current directory
+        alisa count_files="ls -d | wc -l" 
+
+  - bysize => print files ordered by size
+         alisa bysize= "ls -S" 
+
+  - ff => search a file by partial name. For example running "ff ba" should print all files with "ba" in their name
+        alisa ff = "ls -l | find . type f $1,$2
```
ynonp, line 28: I don't think it works. $1, $2 don't mean anything here, `type` requires a dash (i.e. `-type`) and you can't pipe into find.

A working solution combines find and grep:

alias ff="find . -type f | grep "

It's not perfect since it'll also match files that are in a directory that contains the required word, for example searching for ba will also find the file: `bad/file.txt`

Bash functions are a way to create more complex aliases that uses parameters they received in creative ways. For the above example another solution using functions:

```
 function ff2 { find . -name "*$1*"; }
```

Calling the function looks just like running an alias: `ff2 unix`

37_environment_lab/envlab.txt
```diff
@@ -1,15 +1,37 @@
...
+
+  - lsf => like "ls -l" but print only files
+    alisa lsf ="ls -f"
+
+  - lsd => like "ls -l" but print only directories
+   alisa lsd="ls -d"
+
+  - lsg => like "ls -l" but print all directories first and then files
+   alisa lsg="ls -d | ls -f"
```
ynonp, line 37: Combining aliases is performed with `;` (not with `|`), so solution should be: `alisa lsg="lsf; lsd"`

ynonp (commented):
Please review suggestions and comments

PR #164, opened 2017-07-02
ToCode Solution: תרגול awk

moshikob99:


32_awk_lab/awklab.txt
```diff
@@ -1,26 +1,55 @@
...
-
-
-
-
+
+
+1. Write an awk one-liner that prints the file "shells" with a blank line after each line
+
+Answer:  
cat /etc/shells | awk '{print$0;print” ”}'
```
ynonp, line 5: Good. Alternatively we can use printf:

```
cat /etc/shells | awk '{ printf("%s\n\n", $0) }'
```

In printf you specify the format of the line, and it automatically "injects" the parameter (in this case $0, the entire line) instead of the %s

32_awk_lab/awklab.txt
```diff
@@ -1,26 +1,55 @@
...
+
+
+
+
+2. Write an awk one-liner that prints the file "shells" with a ">" sign at the beginning of each line
+
+Answer: cat /etc/shells | awk '{print “>” $0}'
```
ynonp, line 12: Very nice. Also with printf it would look like:

```
cat /etc/shells | awk '{ printf("> %s\n", $0) }'
```

32_awk_lab/awklab.txt
```diff
@@ -1,26 +1,55 @@
...
+


+3. Write an awk one-liner to swap the first and last words in the output of "help cd"
+cat /
help cd | awk '{s=$1; $1=$NF; $NF=s}1'?????
+
+
cat /
help cd | awk '{temp=$1; $1=$NF; $NF=temp ; print $0}'
```
ynonp, line 20: 👍

32_awk_lab/awklab.txt
```diff
@@ -1,26 +1,55 @@
...
+Answer: cat /etc/shells | awk '{print “>” $0}'
+
+
+
+


+3. Write an awk one-liner to swap the first and last words in the output of "help cd"
+cat /
help cd | awk '{s=$1; $1=$NF; $NF=s}1'?????
```
ynonp, line 18: You can't use "1" after a command (1 is used instead of a command, as the default command)

32_awk_lab/awklab.txt
```diff
@@ -1,26 +1,55 @@
...
+    media: autoselect
+
+    status: inactive
+
+
+
+Write an awk one liner that splits this file into multiple file, one for each block, so after running the script we'll have a file named "en3" with the lines of the first block, and another file named "p2p0" with the lines of the second block.
+
+
+cat /sbin/ifconfig | awk 'BEGIN{fileName=en3 }{if (==p2p0) fileName=p2p0;print /bin/tcsh>fileName}'
```
ynonp, line 49: 👍

ynonp (commented):
Glad to see you made it!

PR #165, opened 2017-07-04
ToCode Solution: תרגול awk

Jozefff:


32_awk_lab/awklab.txt
```diff
@@ -1,12 +1,17 @@
 1. Write an awk one-liner that prints the file "shells" with a blank line after each line
 Answer: 
 
+awk '1;!(NR%1){print "";}' /etc/shells
```
ynonp, line 4: Works, but you can use a simplified version:

```
awk ' { print; print ""' } ' /etc/shells
```

Or alternatively with printf:

```
awk '{ printf("%s\n\n", $0) }' /etc/shells
```

32_awk_lab/awklab.txt
```diff
@@ -1,12 +1,17 @@
...
 
 2. Write an awk one-liner that prints the file "shells" with a ">" sign at the beginning of each line
 Answer: 
 
+awk '{print ">"$0}' /etc/shell
+
 3. Write an awk one-liner to swap the first and last words in the output of "help cd"
 
+help cd | awk '{v1=$1;v2=$NF;$1=$NF="";print v2, $0, v1}'
```
ynonp, line 13: Why not substitute in-place?

```
help cd | awk '{ temp=$1; $1=$NF; $NF=temp; print }'
```

32_awk_lab/awklab.txt
```diff
@@ -20,6 +25,50 @@ p2p0: flags=8843 mtu 2304
...
+But looking for opportunity to cut to files by delimiters cant use it.
+yosef@yosef-desktop:~$ /sbin/ifconfig | awk 'BEGIN{file = 0; filename = "ouroutput_" file ".txt"} /Link e*/ {getline; file ++; filename = "ouroutput_" file ".txt"}{print $0 > filename}'
+
+this command cuts into files but THE WHOLE LINE that contains delimeter is omitted
+
+:(
+
+please help to touch up it.
+and if you have useful resourses to start with awk because i don't understand how to use it up to level of this task
+
```
ynonp, line 72: The idea is the command print can take a file name and print the line into the file, for example this will print the first word of each line into a file named output:

```
awk '{ print $1 > output }'
```

Now we can approach the block question. As the beginning of the block is identified by lines that start with a non-space character, let's first print only the first line in each block:

```
/sbin/ifconfig | awk '/^[a-z] { print "New Block:" $0 }' 
```

Now the next step is to take the first word AS the file name, and print to that file:

```
/sbin/ifconfig | awk '/^[a-z] { fname = $1 } { print > fname }' 
```

Each line is printed to the file, and when new blocks are encountered we just change fname.

Hope that helps, or email me for any further questions.

ynonp (commented):
Great work. See my comment re. the last exercise

